PR #34, opened 2025-09-26
feat: add native file sharing

telestrial:
https://github.com/user-attachments/assets/c0a4ca71-dc71-4286-8d73-5e4c0c20d98a

Two types of sharing may come to mind here:

1. Sharing a link to a file on the Sia network. I went out and tried to get a feel for this kind of feature in other apps. I found broad support for what we already do--copy the URL to the clipboard and throw a toast-like notification to the user. If, in the future, we support expiry or links that are encrypted until they have the key, we could expand out into some other custom view where settings are modified.
2. Sharing a *file* in the Sia mobile app via the native iOS share sheet - this is now supported.

I also moved the copy link into the actions menu. I think that's where a user expects it to be.

### DIFF
--- a/src/components/FileActionsSheet.tsx
+++ b/src/components/FileActionsSheet.tsx
@@ -29,7 +29,6 @@ import {
   updateFilePinnedObjects,
 } from '../stores/files'
 import Share from 'react-native-share'
-
 type Props = NativeStackScreenProps<MainStackParamList, 'FileDetail'>
 
 export function FileActionsSheet({ route, navigation }: Props) {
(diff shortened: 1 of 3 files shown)


src/components/FileActionsSheet.tsx
```diff
@@ -59,15 +60,34 @@ export function FileActionsSheet({ route, navigation }: Props) {
...
+    try {
+      await Share.open({
+        url: status.cachedUri,
+        type: file.fileType,
+        filename: file.fileName ?? undefined,
+        subject: `Sia Mobile - ${file.fileType}`,
+      })
+    } catch (e) {
+      if (typeof e === 'string' && !e.includes('User did not share')) {
+        console.warn('File sharing failed:', e)
```
alexfreska: use logger.log for all console logging so that users that hit an error can pop open the log viewer and share the error with us.

alexfreska (approved):
Left one minor comment, looks great!!

PR #68, opened 2025-10-01
feat: add sorting and filtering

telestrial:
[Screen Recording 2025-10-01 at 1.12.56 PM.mov <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.dev/user-attachments/video/d3195682-fe1c-47f6-aaa9-99e0d52745d0.mov)

Adds these features to the Gallery and List views. At this point, we could in theory deleted `useFileList` because `useOrderedFileList` can take no parameters and does the same, but I didn't want to pull that trigger myself. I also of course welcome any naming, organization, or style suggestions. On style, I see this space between the header menu and gallery/list view where I'm putting these two buttons as changing quite rapidly as we add search and make bigger decisions about how this should, so I kept this styling pretty basic for now.

One more thing: I couldn't properly get a "file" in here to test that filter functionality. I assume it works but I was getting invalid signature stuff from the sdk. I at least can tell that the file filter is exclusive to the other categories.

### DIFF
--- a/src/components/FileFilter.tsx
+++ b/src/components/FileFilter.tsx
@@ -1,0 +1,97 @@
+import React, { useMemo, useState } from 'react'
+import { View, Text, Pressable, StyleSheet } from 'react-native'
+import { Square, CheckSquare } from 'lucide-react-native'
+import { type Category, useFilesView } from '../stores/files'
+import ActionSheet from '../components/ActionSheet'
+
+const CATEGORIES: Category[] = ['Video', 'Image', 'Audio', 'Files']
+
+export function FileFilter(): React.ReactElement {
+  const [open, setOpen] = useState(false)
+  const { selectedCategories, toggleCategory, clearCategories } = useFilesView()
+
+  const pressableText = useMemo(() => {
+    const n = selectedCategories.size
+    return n ? `Filter • ${n}` : 'Filter'
+  }, [selectedCategories])
+
+  const Row = ({ cat }: { cat: Category }) => {
+    const checked = selectedCategories.has(cat)
+    const Icon = checked ? CheckSquare : Square
+    return (
+      <Pressable
+        style={styles.modalRow}
+        onPress={() => toggleCategory(cat)}
+        accessibilityRole="button"
+        accessibilityLabel={checked ? `Unselect ${cat}` : `Select ${cat}`}
+      >
+        <Text style={styles.label}>{cat}</Text>
+        <Icon size={20} />
+      </Pressable>
+    )
+  }
+
+  return (
+    <>
+      <Pressable
+        style={styles.pressable}
+        onPress={() => setOpen(true)}
+        accessibilityRole="button"
+      >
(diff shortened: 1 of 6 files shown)


telestrial:
* **#68** <a href="https://app.graphite.dev/github/pr/SiaFoundation/sia-mobile/68?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/SiaFoundation/sia-mobile/68?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

alexfreska:
@telestrial  also hit that problem adding files into the simulator, the workaround I found is to open a file in the browser from a website or you can serve it on your machine (eg 127.0.0.1:8000/file.pdf) and save it to the device with the "Save to Files" option.

src/components/FileFilter.tsx
```diff
@@ -0,0 +1,127 @@
...
+    <>
+      <Pressable
+        style={styles.pressable}
+        onPress={() => setOpen(true)}
+        accessibilityRole="button"
+      >
+        <Text style={styles.pressableText}>{pressableText}</Text>
+      </Pressable>
+
+      <Modal
```
alexfreska: Could use the ActionSheet component here (used in FileActionsSheet), that would sync up the animations and backdrop etc.

src/stores/files.ts
```diff
@@ -77,6 +78,57 @@ export async function readAllFileRecords(): Promise<FileRecord[]> {
...
+): Promise<FileRecord[]> {
+  const { sortBy = 'DATE', sortDir, categories = [] } = opts ?? {}
+  const dir: SortDir = sortDir ?? (sortBy === 'NAME' ? 'ASC' : 'DESC')
+
+  const mimes = categories.length
+    ? Array.from(new Set(categories.flatMap((c) => CATEGORY_TO_MIMES[c] ?? [])))
+    : []
+
+  const where = mimes.length
+    ? `WHERE fileType IN (${mimes.map(() => '?').join(',')})`
```
alexfreska: A simpler way to do this that would also mean we don't have to have exhaustive lists in CATEGORY_TO_MIMES is to just match the prefix for image video and audio, eg:
```
SELECT * FROM files WHERE fileType LIKE 'image/%';
SELECT * FROM files WHERE fileType LIKE 'video/%';
SELECT * FROM files WHERE fileType LIKE 'audio/%';
```

alexfreska (commented):
Looks great, left two comments. Agree on the styling, we can meet soon and figure out how we want to further refine things 👍

src/components/FileFilter.tsx
telestrial: Great idea. Got it!

src/stores/files.ts
telestrial: Yeah! Very nice find. It's in.

alexfreska (approved):

PR #83, opened 2025-10-06
feat: add infinite scrolling

telestrial:


telestrial:
* **#83** <a href="https://app.graphite.dev/github/pr/SiaFoundation/sia-mobile/83?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/SiaFoundation/sia-mobile/83?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

alexfreska (commented):
I took a look and played around with this, here are some notes:
- The triggerChange issue stems from the fact that useSWRInfinite uses special cache keys for its pages (adds an `$inf$` prefix) so our helper is not matching. There are a few more necessary tweaks to get it working but I think we probably should not use this hook so I won't expand.
- useSWRInfinite appends pages of data sequentially, what we really want is to only have the relevant window of data in memory. eg: imagine the user is at a scroll position half way through a list of 10,000 files, useSWRInfinite could only get to this position sequentially and would retain all the pages of data, in addition we want the user to be able to open the app directly into or back into a state like this. eg: open photos scrolled to year 2023.

alexfreska:
On second thought lets merge this pattern first and we can figure out windowing afterwards.

src/stores/files.ts
```diff
@@ -313,3 +358,22 @@ export const useFilesView = create<FilesViewState>((set, get) => ({
   },
   clearCategories: () => set({ selectedCategories: new Set() }),
 }))
+
+// File List Revision store
```
alexfreska: This works great but resets the whole infinite state, shared code for that mutate pattern that keeps the current position while revalidating all pages.

src/components/Gallery.tsx
```diff
@@ -9,10 +10,22 @@ type Props = {
...
+  const handleEndReached = useCallback(() => {
+    if (!isLoadingMore && hasMore) setSize(size + 1)
+  }, [isLoadingMore, hasMore, setSize, size])
+
+  const handleRefresh = useCallback(() => {
+    setSize(1)
+  }, [setSize])
+
   return (
     <FlatList
```
alexfreska, line 25: If I scroll to the bottom I see the loader for a moment, I wonder what the main delay is, maybe we can investigate what the cause is and how to smooth it out.

src/components/Gallery.tsx
```diff
@@ -15,10 +16,22 @@ export function Gallery({
...
+  const isRefreshing = !!files && isValidating && size === 1
+  const isLoadingMore = !!files && isValidating && hasMore
+
+  const handleEndReached = useCallback(() => {
+    if (!isLoadingMore && hasMore) setSize(size + 1)
+  }, [isLoadingMore, hasMore, setSize, size])
+
+  const handleRefresh = useCallback(() => {
+    setSize(1)
+  }, [setSize])
```
alexfreska: Should we move all this into the hook? is it the same in both views?

telestrial: It is. Made this a hook. I think if we refine back to useSwr(), we'll probably move away from this hook being useful/needing to be fundamentally change but that's all part of the incremental process.

src/components/Gallery.tsx
alexfreska: I found the issue, its with GalleryItem reading the fs. I'll fix separately.

alexfreska (approved):

PR #90, opened 2025-10-08
feat: account info

alexfreska:
- Updates the sdk.
- Displays the account info in indexer settings screen, also displays the account key in advanced settings screen.
- Fixes the byte display helper.

### DIFF
--- a/src/screens/SettingsIndexerScreen.tsx
+++ b/src/screens/SettingsIndexerScreen.tsx
@@ -1,5 +1,5 @@
 import { View, Text, StyleSheet } from 'react-native'
-import { colors, palette } from '../styles/colors'
+import { palette } from '../styles/colors'
 import { useIsConnected } from '../stores/sdk'
 import { DotIcon } from 'lucide-react-native'
 import { RowGroup } from '../components/Group'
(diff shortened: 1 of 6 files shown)


alexfreska:
* **#90** <a href="https://app.graphite.dev/github/pr/SiaFoundation/sia-mobile/90?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/SiaFoundation/sia-mobile/90?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/screens/SettingsIndexerScreen.tsx
```diff
@@ -42,6 +48,26 @@ export function SettingsIndexerScreen() {
...
+              <LabeledValueRow
+                label="Account Key"
+                value={account.data.accountKey}
+              />
+              <LabeledValueRow
+                label="Description"
+                value={account.data.description}
+              />
+              <LabeledValueRow
+                label="Pinned Data"
```
telestrial: For this and Max Pinned Data, we might want to start thinking about the commercial representation here. Would something like `Used Storage` and `Storage Limit` or similar make more sense and be more understood by the average user?

telestrial (approved):
Very small comment just about data representation (and it doesn't necessarily only apply here, so maybe something to kick down the road for another time) but otherwise looks good!

src/screens/SettingsIndexerScreen.tsx
n8mgr: We should absolutely pick more user accessible terms here

alexfreska: Good point, I'll update that.

PR #163, opened 2025-10-28
fix: upload progress encoded size

alexfreska:
- Add encodeSize method to calculate total size with erasure coding parameters.

### DIFF
--- a/src/managers/uploadToIndexer.ts
+++ b/src/managers/uploadToIndexer.ts
@@ -1,4 +1,4 @@
-import { Sdk } from 'react-native-sia'
+import { Sdk, encodedSize } from 'react-native-sia'
 import { updateUploadProgress } from '../stores/uploads'
 import { encodeFileMetadata } from '../encoding/fileMetadata'
 import { logger } from '../lib/logger'
@@ -25,6 +25,12 @@ export async function uploadToIndexer(params: {
     return
   }
 
+  const totalEncodedSize = encodedSize(
+    BigInt(data.byteLength),
+    UPLOAD_DATA_SHARDS,
+    UPLOAD_PARITY_SHARDS
+  )
+
   const upload = await sdk.upload({
     maxInflight: UPLOAD_MAX_INFLIGHT,
     dataShards: UPLOAD_DATA_SHARDS,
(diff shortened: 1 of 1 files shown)


alexfreska:
* **#163** <a href="https://app.graphite.dev/github/pr/SiaFoundation/sia-mobile/163?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/SiaFoundation/sia-mobile/163?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/managers/uploadToIndexer.ts
```diff
@@ -34,14 +34,19 @@ export async function uploadToIndexer(params: {
       fileSize: data.byteLength,
     }),
     progressCallback: {
-      progress: (uploaded, encodedSize) => {
+      progress: (uploaded, rawEncodedSize) => {
+        const encodedSizeWithErasure = encodedSize(
```
n8mgr: This is incorrect. The second parameter is already the encodedSize. You can't calculate the final size in the callback. You need to get the actual size of the file you've chosen to upload, call encodedSize with that, then use that value when calculating the percentage instead of the second parameter.

Only use the second parameter when you don't know the size of the file ahead of time, but it's impossible to prevent jumps in that scenario.

src/managers/uploadToIndexer.ts
n8mgr, line 19: Not part of this PR, but why not use a readable stream here so you're not reading the full file into memory?

n8mgr: https://github.com/SiaFoundation/sia-mobile/issues/164

src/managers/uploadToIndexer.ts
alexfreska: 👍🏼 updated

src/managers/uploadToIndexer.ts
alexfreska: 👍🏼 I'll take a look at that

n8mgr (approved):

PR #220, opened 2025-11-14
fix: reset zoom on file detail nav

telestrial:
This resets the image zoom position on file detail nav. It fixes a reported issue. That said, I think the issue creates some questions around the best way to do full screen in the file viewer. Which types? When? How do users get in and out of that. In any case, I consider this a stopgap fix while I examine other approaches across the Fileviewer.

### DIFF
--- a/src/components/MediaConsumers/ImageViewer.tsx
+++ b/src/components/MediaConsumers/ImageViewer.tsx
@@ -1,4 +1,11 @@
-import { useEffect, useRef, useState } from 'react'
+import {
+  useCallback,
+  useEffect,
+  useImperativeHandle,
+  useRef,
+  useState,
+  type Ref,
+} from 'react'
 import {
   ScrollView,
   Image,
@@ -9,13 +16,17 @@ import {
   Platform,
 } from 'react-native'
 
-export function ImageViewer({
-  uri,
-  style,
-}: {
+export type ImageViewerHandle = {
+  resetZoom: () => void
+}
+
+type Props = {
   uri: string
   style?: ViewStyle
-}) {
+  ref?: Ref<ImageViewerHandle>
+}
+
+export function ImageViewer({ uri, style, ref }: Props) {
   const scrollRef = useRef<ScrollView>(null)
   const [size, setSize] = useState({ w: 0, h: 0 })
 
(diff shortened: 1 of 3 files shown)


telestrial:
* **#220** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/220?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/220?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/components/MediaConsumers/ImageViewer.tsx
```diff
@@ -9,65 +16,82 @@ import {
...
-}) {
-  const scrollRef = useRef<ScrollView>(null)
-  const [size, setSize] = useState({ w: 0, h: 0 })
+}
 
-  const onLayout = (e: LayoutChangeEvent) => {
-    const { width, height } = e.nativeEvent.layout
-    setSize({ w: width, h: height })
-  }
+export const ImageViewer = forwardRef<ImageViewerHandle, Props>(
```
alexfreska: Do we still need forwardRef in react 19?

alexfreska:
@telestrial whats the reason it has to reset zoom when toggling to a different component? If the zoom is on the image view why does it affect the outer file details view?

telestrial:
@alexfreska This is a bit of a rabbit-hole but I suspect what's happening has to do with `UIScrollView`: https://developer.apple.com/documentation/uikit/uiscrollview

In short, I think as an optimization React native reuses the `ScrollView` (UIScrollView) we zoom in on in ImageViewer. I think this might all work a lot better if we had a Stack here, consisting of FileViewer and FileDetails. Or, another idea I'm thinking, why not just have a longer single view--this would be similar to the Photos app in that you have your image/video, the header, and the footer and, if you scroll, you see more details about the file. If not, one tap to hide header and the bottom "top" of the details view, zooming, etc. The wrinkle in this fold is the action bar and how that works in these contexts.

I think basically we're just hitting an optimization here that happens to fall badly in this specific use case.

src/components/MediaConsumers/ImageViewer.tsx
telestrial: yes. Great catch!

PR #230, opened 2025-11-17
feat: add horizontal swiping through media

### LINKED ISSUE #39: Ability to swipe through files from the detail viewer
- Ability to swipe to the previous and next file from the viewer.
- The sequence should be based on the current filters and sorting (coming soon).

telestrial:
[Screen Recording 2025-11-17 at 1.58.13 PM.mov <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.com/user-attachments/video/626944f7-f693-44b9-9d12-cdae73b3542a.mov)

This also triggers the infinite scroll additional fetches as needed. Also solves the zoom issue in #220, so, with this, we can close that other one.

Closes https://github.com/SiaFoundation/sia-mobile/issues/39.

### DIFF
--- a/src/components/FileActionsSheet.tsx
+++ b/src/components/FileActionsSheet.tsx
@@ -1,6 +1,6 @@
+import { useCallback } from 'react'
 import { type NativeStackScreenProps } from '@react-navigation/native-stack'
 import { type MainStackParamList } from '../stacks/types'
-import { useCallback } from 'react'
 import {
   Trash2Icon,
   CloudOffIcon,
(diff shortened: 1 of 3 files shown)


telestrial:
* **#230** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/230?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/230?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/components/FileActionsSheet.tsx
```diff
@@ -31,20 +31,23 @@ import {
...
+  fileId?: string
 }
 
 export function FileActionsSheet({
   route,
   navigation,
   sheetName = 'fileActions',
+  fileId,
 }: Props) {
+  const targetFileID = fileId ?? route.params.id
```
alexfreska: Could we get rid of route and pass the id for all call sites?

telestrial: Got it!

src/screens/FileDetailScreen.tsx
```diff
@@ -1,63 +1,173 @@
...
+    setSize,
+    isValidating,
+    hasMore,
+  })
+
+  const activeFileFromList = useMemo(
+    () => files.find((item) => item.id === activeFileID),
+    [files, activeFileID]
+  )
+  const activeFile = activeFileDetails ?? activeFileFromList
```
alexfreska: When is activeFileDetails null?

alexfreska (approved):
One comment, is that extra activeFile lookup necessary? Besides that looks good. Could you confirm it works well on Android, and that swiping in both directions works.

src/screens/FileDetailScreen.tsx
telestrial: This is a good point. By the time we're here, we should expect useFileList to be returning the FileRecord we want to view--that's how we got here. I've gone ahead and adjusted.

PR #232, opened 2025-11-18
feat: add filesheet long press to gallery and filelist

telestrial:
[Screen Recording 2025-11-18 at 10.36.17 AM.mov <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.com/user-attachments/video/368d08a4-a4f2-408e-840d-d14bb4258cd9.mov)

Fairly straightforward. Took a bit of prop drilling but I think that's desirable here because we wouldn't want two implementations of this--one on FileList and one on Gallery. So we unify on the LibraryScreen. Removed a host of unused imports on FileActionSheet, while I was there. No change in functionality to that--just unused stuff hanging around!

Closes https://github.com/SiaFoundation/sia-mobile/issues/140

PR #234, opened 2025-11-18
refactor: use more direct and sychronous hasOnboarded get and set

### LINKED ISSUE #227: Restarting app often returns user to onboarding screen

telestrial:
[Screen Recording 2025-11-18 at 1.31.30 PM.mov <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.com/user-attachments/video/7759f2ba-e218-46ab-8d8c-52ae60e0aaaa.mov)

I don't know that this is necessary, so feel free to suggest we not merge this. 

My thought is that if there is anything going sideways with SWR--some intermediate state and/or mutation/key issue, this cuts out that complexity in favor of a zustand store value that we must make sure stays in sync. It's definitely a trade-off, but I like it because it eliminates some library-side complexity with SWR in favor of another library that seems less complex with zustand. I think this also makes us more directly synchronous for this value.

One danger after this PR would be if someone calls `setSecureStoreHasOnboarded` on its own instead of using setHasOnboardedStatus in `app.ts`. We would get out of sync, but our surface area is low--the "Upload Files" button on the `FinishedOnboardingScreen` and the `Reset App` button.

Closes https://github.com/SiaFoundation/sia-mobile/issues/227 

We can open a new ticket if we see it again.

### DIFF
--- a/src/Root.tsx
+++ b/src/Root.tsx
@@ -7,12 +7,16 @@ import {
 } from '@react-navigation/native'
 import { SafeAreaProvider, SafeAreaView } from 'react-native-safe-area-context'
 import { ToastProvider } from './lib/toastContext'
-import { initApp, shutdownApp, useShowSplash } from './stores/app'
+import {
+  initApp,
+  shutdownApp,
+  useHasOnboardedStatus,
+  useShowSplash,
+} from './stores/app'
 import useLinkedURL from './hooks/useLinkedURL'
 import { useReconnectIndexer } from './hooks/useReconnectIndexer'
 import { RootTabs } from './stacks/RootTabs'
 import { uniqueId } from './lib/uniqueId'
-import { useHasOnboarded } from './stores/settings'
 import { ShareIntentProvider } from 'expo-share-intent'
 import { ShareIntentConsumer } from './components/ShareIntentConsumer'
 import { AppSplash } from './components/AppSplash'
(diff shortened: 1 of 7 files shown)


telestrial:
* **#234** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/234?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/234?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#232** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/232?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

alexfreska:
This issue is that the actual value coming out of secure storage is false so duplicating it to the other state manager on init wont solve it. The secure storage getter is returning its default value, it must be due to an error thrown when its called too quickly during app init. The clearest solution I see is to modify the secure store helpers not to return the default on error, and to retry a few times with backoff, that way we don't quietly get the wrong value. This solution will also prevent similar issues if we ever call other secure storage getters right at init, and also makes the errors more apparent for debugging.

alexfreska:
eg: https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/239. I'm interested to see what the logged error is.

PR #262, opened 2025-12-01
refactor: use non-deprecated VideoView fullscreen props

telestrial:
This is very minor but we were using the `allowFullscreen` prop in `VideoPlayer`, and that prop will be deprecated at some point. I've switched it out for the dev's preferred prop and set explicit values (that are the defaults) just to keep this behavior up to date and more resilient to any changes should we update, etc. Tested on iOS and Android and seems to have no behavior change, which is the desired result.

### DIFF
--- a/src/components/MediaConsumers/VideoPlayer.tsx
+++ b/src/components/MediaConsumers/VideoPlayer.tsx
@@ -39,7 +39,11 @@ export function VideoPlayer({
         ref={videoRef}
         style={StyleSheet.absoluteFill}
         player={player}
-        allowsFullscreen
+        fullscreenOptions={{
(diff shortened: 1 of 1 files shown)


telestrial:
* **#262** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/262?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/262?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

alexfreska (approved):

PR #267, opened 2025-12-12
feat: add useFileViewerWindow hook

telestrial:
This is not plugged in anywhere and shouldn't affect the app at all. I have a new fileviewer MVP I've been using to test it and I think I'm happy with this. Was able to rectify rendering issues I had discussed via DM. It was actually just a bug in here about ordering, but it is now resolved. Essentially, for the item(s) "back" from the currently active, the tiebreaker was in the wrong direction.

PR #277, opened 2025-12-23
refactor: add FileCarousel components and  decouple from gallery/filelist

telestrial:
`FileCarousel` replaces `FileDetailScreen` and the like. It is a detached component that accepts a starting `FileRecord` and then utilizes the hook down stack `useFileCarouselWindow` to move about the current "file list", based on the current sorting/filter/etc. This is a decently large edit, affecting several different components. There's been a particular focus to try to optimize this. I think it is unwise to optimize unnecessarily but, this being a sort of second pass at this functionality, one thing I wanted to ensure is that we made the most of the db calls we were making or that were local to the component--specifically around re-renders on data that the user would be currently viewing. I think these are the two big optimizations on that front:

1. When we invoke `FileCarousel`, we already have the exact FileRecord we want to display. We should avoid taking something like the ID alone and then going back to the DB to get that information a second time. We should just display that already fetched data and let the SWR hook go and get the current window, which will include that `FileRecord`.

2. When we swipe across the information provided in `useFileCarouselWindow`, we should ignore the second fetch in favor of the first. So, files 1 2 3. on 1. We swipe right to get to 2. Our window already gave us the FileRecord for 2, so we should resist any kind of render around replacing data that is in the user's view. Let me know if you need more information about that, but the idea is that we only re-render on the "same" file if that file has changed (it's updatedAt value is different, specifically).

I think there probably are likely to be some bugs, this being a pretty large change, in terms of a core functionality of the app. ~~We also don't have a share URL good case to test here as that's currently being reworked.~~ But I'm happy with where it's at and haven't found any stable issues that I can exactly pinpoint. I think we run with it and respond to any found issues.


telestrial:
* **#289** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/289?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#288** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/288?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#287** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/287?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#286** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/286?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#284** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/284?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#282** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/282?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#277** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/277?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/277?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#267** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/267?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#281** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/281?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/components/FileCarousel/index.tsx
```diff
@@ -0,0 +1,557 @@
...
+  const shouldIgnoreNextToggleRef = useRef(false)
+  const imageZoomRef = useRef({ isZoomed: false })
+  const screenWidth = Dimensions.get('window').width
+
+  // Create fresh Animated.Value when currentFile changes to avoid stale native driver values
+  const translateX = useMemo(() => {
+    return new Animated.Value(0)
+  }, [currentFile?.id])
+
+  // Initialize currentFile from initial data and handle SWR revalidation
```
alexfreska: Maybe we can tuck all this file list and prev / next logic into useFileCarouselWindow and return the data?

src/components/FileCarousel/index.tsx
```diff
@@ -0,0 +1,557 @@
...
+            navigation={navigationProxy}
+            icon="close"
+          />
+        </View>
+      ) : null}
+
+      {currentFile ? (
+        viewStyle === 'consume' ? (
+          <View style={styles.viewer} pointerEvents="box-none">
+            {renderFileViewer(prevFile, 'prev')}
```
alexfreska, line 358: If this is hardcoded with just prev/next we cant ever adjust neighbors per side above 1 without refactoring.

src/components/FileCarousel/index.tsx
```diff
@@ -0,0 +1,557 @@
...
+            {renderFileViewer(currentFile, 'current')}
```
alexfreska, line 359: Right now only the current file gets auto downloaded, this means when we swipe to a file that is not local yet the download starts only when the swipe is complete. Could we turn this into a component and call the auto download check, that way we auto download every file / neighbor in the current window.

alexfreska (commented):
Looking good, left a few comments!

src/components/FileCarousel/index.tsx
telestrial: Done downstack.

src/components/FileCarousel/index.tsx
telestrial: Handled. I needed to introduce a new component: `FileCarouselPage`, but that's fine.

src/components/FileCarousel/index.tsx
telestrial: This one I'm a bit less apt to do at the moment. I agree that, if we wanted to extend the window, that would require an edit to FileCarousel, but, right now, that doesn't seem like something we want to do. So, I'd be adding logic to handle that dynamic nature and possibly some additional state and, after that, no functionality would change. I think I'd rather add this when/if we need it.

alexfreska (approved):

telestrial:
### Merge activity

* **Jan 7, 5:10 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/277).
* **Jan 7, 5:14 PM UTC**: [Graphite](https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/277) rebased this pull request as part of a merge.
* **Jan 7, 5:14 PM UTC**: @telestrial merged this pull request with [Graphite](https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/277).

PR #316, opened 2026-01-22
feat: add sync event handling to FileCarousel

telestrial:
This PR takes some opinionated actions based on syncDown events. I've defined a few scenarios:

1.  **The file in view is deleted** - We inform the user by toast and close their view. This is probably the most opinionated, and I did consider letting the user choose. There's some questions here around "what does it mean if a user deletes a file *from the network* on device? Does that mean all devices should remove it?" I came to a yes on that, but obviously this or any other way I've come down on these can change.

2. **Files before or after are added or removed** - I decided this should be imperceptible to the user. But this does mean that you might "miss" a file behind you. So if you were sorted by name and on B and there was a file add that was the last A, you're not going to see that come in behind you. But it will appear when you swipe back. You won't go to where you were before--the now second-to-last A. And the same with deletions.

3. **File updated** - We update the file before your eyes and let you know that we did. There is not much here that would actually be visible on the screen, I don't think--maybe just the name in the header, if it's visible, but I think letting them know what they're looking at just updated is important.

4. **Current file's position changed** - this is mostly about file names. If you rename a file on a new device, we update it in front of you on the other device, toasting that it's changed as per 2 above, and now that file sits somewhere else. We invalidate our cache basically and re-fetch neighbors, etc. This would probably be a touch disorienting because, from the user's perspective, they were at item 8 of 10, maybe, and suddenly the file they're looking at is 4. BUT...we do inform them, so that should quash that confusion. We even say, in this case, "File name has changed." or something similar.

### DIFF
--- a/src/stores/fileCarousel.ts
+++ b/src/stores/fileCarousel.ts
@@ -9,6 +9,7 @@ import {
   Category,
   useLibrary,
   buildLibraryQueryParts,
+  librarySwr,
 } from './library'
 
 const FILE_COLUMNS =
@@ -232,11 +233,34 @@ export async function fetchFilesAtIndices(
 
 // The carousel can't hold thousands of files in memory, so we give it placeholder
 // indices and fetch the actual file data as the user swipes nearby.
+
+async function fetchFileByID(fileID: string): Promise<FileRecord | null> {
+  const row = await db().getFirstAsync<FileRecordRow>(
+    `SELECT ${FILE_COLUMNS} FROM files f WHERE f.id = ? LIMIT 1`,
+    fileID
+  )
+
+  if (!row) return null
+
+  const [file] = await hydrateRows([row])
+  return file ?? null
+}
+
+async function fileExists(fileID: string): Promise<boolean> {
+  const result = await db().getFirstAsync<{ count: number }>(
+    `SELECT COUNT(*) as count FROM files WHERE id = ?`,
+    fileID
+  )
+  return (result?.count ?? 0) > 0
+}
+
 type UseVirtualFileListParams = {
   initialId: string
   initialFile?: FileRecord
   prefetchRadius?: number
   maxCacheSize?: number
+  onDeleted?: () => void
+  onUpdated?: (message: string) => void
 }
 
 type UseVirtualFileListReturn = {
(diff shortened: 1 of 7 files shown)


telestrial:
* **#319** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/319?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#318** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/318?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#317** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/317?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#316** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/316?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/316?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#315** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/315?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/stores/fileCarousel.ts
```diff
@@ -246,6 +270,8 @@ export function useVirtualFileList({
   initialFile,
   prefetchRadius = 3,
   maxCacheSize = 50,
+  onDeleted,
+  onUpdated,
```
alexfreska, line 293: It would valuable to have the behaviours described on your PR description encoded in some tests against this hook. Not sure we have an example of a test harnessing a React hook (vs function) yet but it would be very valuable to set up since this carousel logic is a complicated and important.

alexfreska (approved):
Regarding the list of scenarios, I agree with each decision. #1 is totally correct, any decisions on keeping the file around should occur at the sync layer.  
  
It would be great to add some tests, let me know what you think on that or if you decide to and have trouble setting it up.

src/stores/fileCarousel.ts
telestrial: Went ahead and added `testing-library/react-native`, which gives us a native way to test hooks, and then wrote those hook tests.

telestrial:
### Merge activity

* **Jan 28, 6:21 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/316).
* **Jan 28, 6:22 PM UTC**: [Graphite](https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/316) rebased this pull request as part of a merge.
* **Jan 28, 6:23 PM UTC**: @telestrial merged this pull request with [Graphite](https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/316).

Issue #358, opened 2026-02-03
File carousel viewer shard flickering


alexfreska:
- When viewing a text file, after the file and viewer has loaded, sometimes the loading state will re-run. I am guessing this is when the library/db receives updates. Reported here: https://siafoundation.slack.com/archives/C09M760RM60/p1770097540069219

alexfreska:
A few related fixes have been merged.

PR #360, opened 2026-02-04
fix: prevent current file flicker when library updates

telestrial:
This addresses https://github.com/SiaFoundation/sia-mobile/issues/358. 

When the total count of the sort/filter list changes, we invalidate the FileCarousel cache. This is necessary because even one file changing could change the current position of tall files, including the one we're currently viewing. Things out of view within our window can re-fetch without the user's knowledge. However, the currently viewed file doesn't need to be evicted—its data is still valid, just at a different index. By preserving and restoring the current file across cache clears, we avoid a flash of loading state while the file is re-fetched. Updates to the currently viewed file are handled separately further down in handleLibraryChange, where fresh data is fetched and the cache is updated in place.

### DIFF
--- a/.changeset/fix_file_viewer_flickering_when_library_updates.md
+++ b/.changeset/fix_file_viewer_flickering_when_library_updates.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+Fix file viewer flickering when library updates.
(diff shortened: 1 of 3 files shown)


telestrial:
* **#360** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/360?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/360?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

alexfreska (approved):
Tested this out, appears to reduce the flickering! I still sometimes get not found state when I open a cloud only file during heavy metadata sync and downloads (usually it does show Download queued). I'll try to take a look at that.

PR #371, opened 2026-02-09
feat: add --release flag to dev builds for standalone phone usage

alexfreska:
- Adds `dev:ios:device:release` and `dev:android:device:release` which puts a sia.storage.dev release build on a device that can be used without a dev server.

PR #372, opened 2026-02-09
feat: rewrite upload system with async loop, slab-aware packing, and pipelined adds

alexfreska:
> TLDR refinements to upload packing from testing on a real device, this PR adds about 2k LOC of tests to cover scenarios and edge cases that were observed.

- Replace the interval-based upload scanner with an UploadManager that runs its own async loop, polling the DB for local-only files and processing an explicit queue for user-initiated uploads. Exclude active upload IDs at the SQL level so the query limit returns actually-available files.
- Files are packed into slab-aligned batches that flush on four triggers: slab fill threshold (≥90%), max slabs (10), max duration (60s with suspension-gap detection), and idle timeout (10s after last file). Each packer.add() call packs data and uploads full slabs to the network as they fill; finalize() handles only the last partial slab.
- Pipeline up to 200 concurrent packer.add() calls to overlap reader creation and FFI call setup with the previous call's pack+upload I/O. Add shutdown safety checks throughout the loop so stopping takes effect immediately.
- Add packed/uploaded counters to UploadManager for accurate background task session logging (replaces unreliable file-count delta which was corrupted by concurrent thumbnail generation). RAF-throttle upload progress updates to avoid excessive store churn.


alexfreska:
* **#376** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/376?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#375** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/375?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#374** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/374?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#373** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/373?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#372** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/372?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/372?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#371** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/371?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/managers/uploader.ts
```diff
@@ -201,251 +489,306 @@ class UploadManager {
...
+    // t1→t2 measures this file's queue wait + processing + any slab upload.
+    for (const { entry, readerMs, promise } of inflight) {
       try {
-        await this.flush('idle_timeout')
-      } catch (error) {
+        const t1 = Date.now()
+        await promise
+        if (!this.active) break
+        const t2 = Date.now()
+        const slabsBefore = this.batch!.slabsFilled
```
telestrial: Is this.batch guaranteed to exist here? If flush() can happen earlier in this loop (and ends the current batch), do we want to keep processing the rest of the window afterward?

src/managers/uploader.ts
```diff
@@ -201,251 +489,306 @@ class UploadManager {
...
+        registerUpload(entry.fileId)
+        this.polledFiles.push(entry)
+      }
 
-      // Match objects to files and save, returns IDs of successfully saved files
-      const successfulFileIds = await this.saveBatchObjects(
-        batch,
-        pinnedObjects,
-      )
+      return this.polledFiles.length
```
telestrial, line 665: How is the return value here meant to be interpreted by the loop? New work? If so, wouldn't this return the count and the loop would go on as if that's new work? Definitely could be missing something with this or the previous comment.

telestrial (approved):

telestrial:
For my two comments and then sort of zooming out a bit: I don’t understand the entire scheme yet since there’s a lot going on here, but it might be worth double-checking some assumptions around what certain signals or return values are meant to represent, and how they’re then used to guide control flow later on.

src/managers/uploader.ts
alexfreska: This is a good catch, I will actually just remove flush from the loop and then we wont have this issue at all. The flush is an over optimization that actually creates more issues than its worth because it leaves the rest of the promises dangling, and theres really no reason not to finish the full batch and go over the max duration a little bit. The max duration does not have to be very strict its just to make sure we flush often enough to not lose upload work.

src/managers/uploader.ts
alexfreska: Yes it's new work. The loop polls the DB, adds the files it finds to the packer (flushing batches as slab/duration limits are hit), then polls again. It keeps going until pollDB returns 0, meaning no more local-only files to upload. Each poll excludes files already being processed via excludeIds so it always gets the next batch, not the same ones.

alexfreska:
### Merge activity

* **Feb 10, 7:55 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/372).
* **Feb 10, 7:58 PM UTC**: [Graphite](https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/372) rebased this pull request as part of a merge.
* **Feb 10, 7:58 PM UTC**: @alexfreska merged this pull request with [Graphite](https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/372).

PR #389, opened 2026-02-13
fix: eliminate filesystem race causing download loop on Android

alexfreska:
- iOS was working fine but it seems like on Android sometimes there is a race condition. This change should completely remove the **possibility** by setting the swr value directly.
- Set the SWR cache directly in copyFileToFs with the known URI instead of re-fetching from the filesystem, where Android's [File.info](http://File.info)() can return exists: false immediately after File.copy().
- Remove the 'done' download status and 5s timer workaround that this makes unnecessary.

### DIFF
--- a/.changeset/fix-android-download-loop.md
+++ b/.changeset/fix-android-download-loop.md
@@ -1,0 +1,5 @@
+---
+default: patch
+---
+
+Fixed a bug on Android where downloaded files would loop back to downloading.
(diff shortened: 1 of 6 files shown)


alexfreska:
* **#389** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/389?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/389?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #393, opened 2026-02-17
fix: properly reset app state and return to onboarding

alexfreska:
Reset app does not always work and often requires restarting the app. This is because services are not cleanly shutdown and therefore can continue to update state, plus a few other caches are not being reset. This PR adds the ability to abort services fully, and makes sure we cover all caches and state.

- Add AbortSignal support to service intervals so in-flight workers (particularly syncDownEvents' while loop) can be stopped during shutdown. 
- Overhaul resetApp to use AsyncStorage.clear() instead of fragile per-key clearing, reset syncDown isSyncing state, and only start sync services after onboarding.
- Trigger syncDownEvents immediately after onboarding completes.

PR #392, opened 2026-02-17
perf: batch DB transactions and add adaptive interval in syncDownEvents

alexfreska:
- This PR makes the sync down process significantly faster. Before it was very serial and slow mostly so things were easier to debug, this moves to large batches that are commited in a single transaction, and also removes waiting between batches. 
- Split processBatch into prepare/commit/cleanup phases so all DB writes in a batch share a single mutex acquire and SQLite transaction. FS cleanup for deletes is now non-fatal (errors logged, cursor still advances). Return 0 from syncDownEvents when multiple events are found to trigger immediate re-polling instead of waiting the default 10s.
- When sync-down-events is actively pulling files, show an ActivityIndicator with "Syncing your files" instead of the "Add files to get started" empty state. Adds a lightweight Zustand store to track sync-down progress.

PR #391, opened 2026-02-17
refactor: fine-grained cache invalidation for SWR layer

alexfreska:
This and the next PR aim to reduce unnecessary re-rendering / revalidation. Right now the swr keys revalidate everything constantly. And the sync service triggers these updates for every event. The changes in this PR also move to setting new values into the swr cache whenever possible rather than triggering a general mutate and letting hooks revalidate which can cause some intermediate state issues.

Replace blanket cache invalidation with targeted updates scoped to exactly what changed. Mutations now only refresh the specific stats, lists, or records they affect instead of triggering full reloads.

- Typed cache factories with auto-generated keys replace shared mutable helpers
- Sync loops batch invalidation once after the loop instead of per-event
- Gallery list revalidates pages in place instead of full key swap
- useFileStatus uses useMemo instead of unnecessary SWR async layer

PR #394, opened 2026-02-18
feat: add metadata migration system and resolve hash uniqueness issues

alexfreska:
We originally used content hash as the unique identifier for files. It makes the sync simple - if two objects had the same hash, they were the same file. But this causes real problems. Two thumbnails that happen to produce the same hash (solid-color images, default placeholders) collide silently - the second gets rejected, the file never gets a thumbnail, and the scanner retries forever. It also means users can't have duplicate files, which they may want intentionally. And the uniqueness constraint hides duplicates from the system entirely - we never know they exist, so we can never clean them up.

Fixing this required two things: a system to safely evolve object metadata, and then using it to fix the schema.

## Versioned metadata encoding

Each object carries metadata as unstructured JSON, stored locally and on the indexer. Until now there was no way to evolve the format without breaking existing accounts.

This is much harder than a typical DB migration. The indexer is the source of truth, but each client has its own database, can be running any schema version, works offline, and reads or writes metadata assuming its own version at any time. There's no atomic "migrate everything" - one device might be mid-upgrade while another is syncing normally. A client might see a mix of v0 and v1 objects in the same sync batch. An old client that doesn't understand v1 could overwrite it with v0 and strip new fields.

The approach is to update the normal sync paths to handle mixed versions gracefully. syncDown accepts both v0 and v1 metadata and normalizes on read. syncUp checks the remote version before writing and refuses to overwrite newer formats. syncUpMetadata drives v0→v1 upgrades incrementally during normal operation. Even across multiple devices at different app versions, the system converges over time.

The infrastructure:

- Object metadata is now versioned with Zod schemas (v0 = old unversioned format, v1 = current). decodeFileMetadata handles any version gracefully, and future versions are parsed with safe defaults so older clients don't break.
- syncUpMetadata refuses to overwrite metadata written by a newer app version, preventing an old client from stripping fields it doesn't understand. When the remote already has a canonical ID set by another device, it's preserved rather than overwritten.
- syncUpMetadata processes batches concurrently (30 slots via SlotPool) and halts cursor advancement on errors so failed files are retried on the next pass.
- Future metadata versions plug into the same encoding and sync paths handle them automatically.

## v1 file identity: schema changes

Using the system above, the first version fixes the root cause - the DB treating content hash as identity:

- kind column - files and thumbnails are now explicitly tagged ('file' / 'thumb') instead of inferred from NULL checks. Orphaned thumbnails stay hidden from the library regardless of column state.
- thumbForId replaces thumbForHash - thumbnails reference their parent by stable ID, not content hash.
- No uniqueness on content - UNIQUE on hash is gone. (thumbForId, thumbSize) is a regular index. Duplicate thumbnails are allowed at the DB level and cleaned up by the application (future PR). No more silent rejections.
- syncDown resolves both files and thumbnails through the same lookup chain: metadata.id → objectId → create new.


alexfreska:
* **#398** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/398?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#397** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/397?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#396** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/396?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#394** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/394?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/394?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#393** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/393?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#392** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/392?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#391** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-mobile/391?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

telestrial (approved):

PR #423, opened 2026-02-27
feat: add trash and soft delete with cross-device sync

### LINKED ISSUE #46: Trash / recently deleted
- Add a soft-delete trash / recently deleted feature.

alexfreska:
- Files are now trashed instead of permanently deleted. The default "Delete" action in the file actions sheet moves files to trash, and the toast says "Moved to trash." Trashing is reversible via restoreFiles.
- Permanent delete sets a deletedAt tombstone on the file row instead of removing it. The tombstone ensures deleted files are never resurrected by sync — even if the device is offline when the delete happens and reconnects later.
- syncUp now handles remote object cleanup for tombstoned files, calling sdk.deleteObject() on each tick instead of making ad-hoc network calls at delete time. This means deletes are retried automatically and survive app restarts.
- syncDown creates tombstones when it receives deleted: true events from the indexer, instead of hard-deleting the file row. The file row persists as a permanent record of deletion.
- trashedAt syncs across devices via object metadata (v1 encoding). Trashing a file on one device propagates to all others through the normal sync cycle.
- Files in trash for more than 30 days are auto-purged to tombstones on app boot.
- Adds migration 0007_soft_delete with trashedAt and deletedAt columns plus indexes.
- Removes the ad-hoc deleteFileFromNetwork, deleteAllIndexerObjects, and "remove from network" UI option — all remote cleanup is now owned by syncUp.
- Closes https://github.com/SiaFoundation/sia-mobile/issues/46

### DIFF
--- a/apps/mobile/src/lib/deleteFile.tsx
+++ b/apps/mobile/src/lib/deleteFile.tsx
@@ -1,48 +1,82 @@
-import { tryCatch } from '@siastorage/core'
-import type { LocalObject } from '@siastorage/core/encoding/localObject'
+import { TRASH_AUTO_PURGE_AGE } from '@siastorage/core/config'
 import { logger } from '@siastorage/logger'
-import {
-  deleteFileRecordAndThumbnails,
-  deleteManyFileRecordsAndThumbnails,
-  type FileRecord,
-} from '../stores/files'
+import { db, withTransactionLock } from '../db'
+import type { FileRecord } from '../stores/files'
+import { readFileRecordsByIds } from '../stores/files'
 import { removeFsFile } from '../stores/fs'
 import {
-  deleteLocalObjects,
-  deleteManyLocalObjects,
-} from '../stores/localObjects'
-import { getSdk } from '../stores/sdk'
+  invalidateCacheLibraryAllStats,
+  invalidateCacheLibraryLists,
+} from '../stores/librarySwr'
 import { removeTempDownloadFile } from '../stores/tempFs'
 import { removeUploads } from '../stores/uploads'
 
-async function tryStep<T>(
-  step: string,
-  id: string,
-  fn: () => Promise<T>,
-): Promise<void> {
-  const [, err] = await tryCatch(fn)
-  if (err) {
-    logger.error('deleteFile', 'step_failed', { step, id, error: err as Error })
-  }
+/**
+ * Move files to trash by setting trashedAt. Trashed files are hidden from
+ * the library but can be restored. The updatedAt timestamp is bumped so
+ * syncUp pushes the change to the indexer. Associated thumbnails are also
+ * trashed.
+ */
+export async function trashFiles(fileIds: string[]): Promise<void> {
+  if (fileIds.length === 0) return
+  const now = Date.now()
+  const placeholders = fileIds.map(() => '?').join(',')
+  await withTransactionLock(async () => {
+    await db().runAsync(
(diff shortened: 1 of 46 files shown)


alexfreska:
* **#428** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/428?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#427** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/427?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#426** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/426?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#425** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/425?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#424** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/424?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#423** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/423?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/423?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

apps/mobile/src/lib/deleteFile.tsx
```diff
@@ -52,64 +86,63 @@ export async function permanentlyDeleteFiles(
...
+/**
+ * Permanently delete files that have been in the trash longer than
+ * TRASH_AUTO_PURGE_AGE (30 days). Sets the deletedAt tombstone so
+ * syncUp will handle remote object cleanup on the next tick.
+ */
+export async function autoPurgeOldTrashedFiles() {
+  try {
+    const cutoff = Date.now() - TRASH_AUTO_PURGE_AGE
+    const rows = await db().getAllAsync<{ id: string }>(
+      `SELECT id FROM files WHERE trashedAt IS NOT NULL AND trashedAt < ? AND deletedAt IS NULL AND kind = 'file'`,
```
telestrial, line 135: Might be the wrong comment location but as good as any, I think: what happens when a file that has been deleted (locally tombstoned) is re-added? Is there any race in terms of syncUp calling sdk.delete and that? or even after both being tombstoned and then being re-added?

alexfreska: If the user adds the same file again, its a new row in the files table with a new ID, it will have the same sha256 hash value as the old row but otherwise all our file identity stuff is now based on the ID and does not enforce unique hash so its treated as a new file with similar metadata - also when the new file uploads the sdk uses a different encryption key so the object IDs will be different too, same story for the thumbnails.  
If the same file is re-added via photos sync feature, it will not get added if there is a matching file in the trash, but it does get added again if the old file is fully deleted/tombstoned because those records are effectively ignored by all queries besides ones for sync logic.

telestrial (approved):

alexfreska:
### Merge activity

* **Mar 3, 2:12 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/423).
* **Mar 3, 2:12 AM UTC**: @alexfreska merged this pull request with [Graphite](https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/423).

PR #483, opened 2026-03-20
fix(sdk-mock): match real indexer objectEvents behavior

alexfreska:
- MockSdk now upserts events per objectId (one event per object, matching the real indexer's unique constraint) and uses keyset cursor pagination (updatedAt ASC, objectKey ASC) matching the real indexer's SQL query.
- It was close the same before but was not exact and had a one small potential edge case that could cause a test to flake.
- Removes deduplicateEvents test helper and fake future timestamps that were workarounds for the old append-only event accumulation.

### DIFF
--- a/apps/integration/test/multi-device-convergence.test.ts
+++ b/apps/integration/test/multi-device-convergence.test.ts
@@ -34,14 +34,6 @@ async function waitForAllObjectsV1(
   )
 }
 
-function deduplicateEvents(storage: MockIndexerStorage): void {
-  const latestByObjectId = new Map<string, (typeof storage.events)[number]>()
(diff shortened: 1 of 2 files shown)


alexfreska:
* **#483** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/483?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/483?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

telestrial (approved):

PR #507, opened 2026-04-01
feat(core): add file versioning support, related perf benchmarks and improvements

alexfreska:
- Files with the same name in the same directory are treated as versions.
- Related perf improvements and benchmark tests.
- Added bulk app service APIs and significantly speed up sync down.


alexfreska:
* **#511** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/511?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#512** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/512?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#501** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/501?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#510** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/510?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#509** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/509?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#508** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/508?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#507** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/507?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/507?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#506** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/506?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#505** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/505?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#515** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/515?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/core/src/db/migrations/0003_version_indexes.ts
```diff
@@ -0,0 +1,24 @@
+import type { DatabaseAdapter } from '../../adapters/db'
+import type { Migration } from '../types'
+
+async function up(db: DatabaseAdapter): Promise<void> {
+  await db.execAsync(
+    `CREATE INDEX IF NOT EXISTS idx_files_version_group
```
telestrial, line 15: Commenting here to express the view, in case you haven't thought about this already, that the NOT EXISTS pattern throughout this PR has performance implications, even with the index, and we'll want to keep an eye on that. I'm thinking about large data sets with many duplicates. I don't know that there is anything to act on right now.

packages/core/src/db/operations/library.ts
```diff
@@ -20,6 +20,25 @@ const MEDIA_PREFIXES: Record<MediaCategory, string> = {
...
+/**
+ * NOT EXISTS subquery that keeps only the latest version per (name, directoryId)
+ * group. A file is "current" if no other active file with the same name and
+ * directory has a higher (updatedAt, id). The partial index
+ * idx_files_version_group makes each probe O(1).
+ */
+export function buildLatestVersionFilter(alias: string): string {
+  return `NOT EXISTS (
+    SELECT 1 FROM files _fv
+    WHERE _fv.name = ${alias}.name COLLATE NOCASE
```
telestrial: User input file names is chaos...much like network requests and the like. I don't know how much meaning we can derive from a user naming something `Foo.txt` vs. `foo.txt`. They could intend for these to be entirely separate entities, and it's really hard to game that out and know. If anything, we should be on the look out for misunderstandings around this.

packages/core/src/db/operations/files.ts
```diff
@@ -782,3 +785,102 @@ export async function deleteLostFiles(
...
+}
+
+export async function trashAllFileVersions(
+  db: DatabaseAdapter,
+  name: string,
+  directoryId: string | null,
+): Promise<string[]> {
+  const versions = await queryFileVersions(db, name, directoryId)
+  if (versions.length === 0) return []
+  const { trashFiles } = await import('./trash')
```
telestrial: You might consider refactoring in a way where dynamic importing like this is not necessary. I can admit I don't fully understand the context but it reads like a hack of a kind. Is there a more idiomatic way to do this?

packages/core/src/db/operations/library.test.ts
```diff
@@ -505,13 +505,13 @@ describe('queryFilePositionInSortedList', () => {
     })
 
     test('breaks NAME ties using ID', async () => {
```
telestrial: Does this test still do what it was originally written to do or have you moved away from that original purpose and, if so, is the case you left behind something you don't need to test any longer?

telestrial (approved):

packages/core/src/db/operations/library.test.ts
alexfreska: Fixed the name

packages/core/src/db/operations/library.ts
alexfreska: Oops this was a mistake, as you probably staw upstack I removed all NOCASE. Updated to remove it here.

packages/core/src/db/operations/files.ts
alexfreska: Oops good catch, removed it, I'm going to try to add a lint rule that errors on these unless they have an explicit comment allowing them.

packages/core/src/db/migrations/0003_version_indexes.ts
alexfreska: Yes this is super important. I benchmarked (~100k records, where 50k are unique the rest are versions on those, some with 2, 3, 4 versions etc) the change  and with the covering index the main queries are fast, the ANALYZE call is very important, without it queries were not automatically finding the index. The aggregation queries get a little slower but still fairly fast and these are for stats and not in the hot path like scrolling through files. That being said, I need to test it some more on real devices.

alexfreska: Added a benchmark app that benchmarks with 1m files. Also adjusted the NOT EXISTS to a current column, as it performed better, tradeoff being requiring updating the column on writes.

alexfreska:
### Merge activity

* **Apr 5, 6:14 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/507).
* **Apr 5, 6:14 PM UTC**: @alexfreska merged this pull request with [Graphite](https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/507).

PR #527, opened 2026-04-08
fix(mobile): release SQLite WAL lock before background suspension

alexfreska:
User reported crashes when backgrounding the app. iOS was killing the process for holding SQLite WAL file locks while suspended.

Adds a suspension manager that signals services to exit via their abort signals, then closes the database before iOS suspends the process. Resuming the app reopens the database and restarts services. The suspension manager handles both normal app backgrounding and background task execution, resuming the database when a task fires and re-suspending when it finishes. The suspension manager was tested and should work even when the user is opening and closing the app frequently.

### DIFF
--- a/.changeset/fix-background-suspension-crash.md
+++ b/.changeset/fix-background-suspension-crash.md
@@ -1,0 +1,5 @@
+---
+mobile: patch
+---
+
+Fixed iOS killing the app when backgrounded by closing the database connection before suspension to release SQLite WAL file locks.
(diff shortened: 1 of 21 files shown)


alexfreska:
* **#527** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/527?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/527?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #653, opened 2026-05-04
fix(mobile,core): centralize AppState reads and treat iOS inactive as foreground

alexfreska:
Goal of this PR is to add some more determinism to the ordering of status change event subscribers, and get rid of any inconsistent states.

- Centralizes iOS foreground/background detection into one lifecycle module so all subsystems (suspension manager, SWR's pause gate, SWR's focus signal) see transitions in a single fixed order instead of each reading AppState independently.
- Reclassifies iOS 'inactive' as a foreground sub-state per Apple's docs, so services keep running through transient interruptions like notification banners and Face ID prompts instead of pausing.

### DIFF
--- a/.changeset/inactive-is-foreground.md
+++ b/.changeset/inactive-is-foreground.md
@@ -1,0 +1,6 @@
+---
+mobile: patch
+core: patch
+---
+
+Treat iOS `inactive` AppState as a foreground sub-state (per Apple's docs) so SWR data fetches keep running through transient interruptions like notification banners and Face ID prompts instead of pausing.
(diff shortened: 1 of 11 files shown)


alexfreska:
* **#653** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/653?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/653?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

telestrial (approved):

PR #659, opened 2026-05-05
fix(mobile): decode percent-encoded file:// URIs before copying imports

alexfreska:
Importing a file with spaces (or any character the OS percent-encodes) silently failed: pickers hand us a file:// URL like Screenshot%202025-09-03.png, but the copy step only stripped the scheme and passed the still-encoded path to RNFS, which couldn't find the real file on disk. The adapter now decodes the path component when converting file:// URIs to filesystem paths, with a try/catch fallback so a malformed % sequence degrades to the old behavior instead of throwing. https://github.com/SiaFoundation/sia-storage-app/issues/610

### DIFF
--- a/.changeset/fix-import-filenames-with-spaces.md
+++ b/.changeset/fix-import-filenames-with-spaces.md
@@ -1,0 +1,5 @@
+---
+mobile: patch
+---
+
+Fixed an issue where importing files with spaces (or other special characters) in their names would fail silently. Closes https://github.com/SiaFoundation/sia-storage-app/issues/610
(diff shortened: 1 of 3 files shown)


alexfreska:
* **#659** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/659?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/659?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

telestrial (approved):

PR #744, opened 2026-06-08
fix(mobile,core): generate thumbnails natively so rotated photos preview upright

alexfreska:
Rotated photos were previewing sideways because the orientation each platform's image decoder applied varied by OS and format, forcing the app to guess the correction from decoded dimensions — fragile and wrong in cases like iPhone HEICs, where the rotation lives in the HEIF container rather than an EXIF tag. This moves the whole thumbnail pipeline into a small native Expo module that decodes at reduced resolution, applies EXIF/HEIF orientation in the same pass, and writes the result straight to disk — so one decoder owns the pixels end to end and the guesswork is gone (also tightening the per-platform list of which file types each OS will attempt).


alexfreska:
* **#744** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/744?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/744?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

telestrial (approved):

PR #746, opened 2026-06-10
feat(mobile): add "Save to device" export option on Android

alexfreska:
Android users had no way to export a file to their device storage: the app's only export path is the system share sheet, which on Android lists only apps — unlike iOS, where the sheet has a built-in "Save to Files" target. The file actions menu now offers "Save to device" on Android (when a local copy exists), which opens the system folder picker and streams the file natively into the chosen destination, handling picker cancellation silently and cleaning up the empty target document if the write fails.

### DIFF
--- a/.changeset/android-save-to-device.md
+++ b/.changeset/android-save-to-device.md
@@ -1,0 +1,5 @@
+---
+mobile: patch
+---
+
+Android now has a "Save to device" option in the file actions menu. It opens the system folder picker so you can choose where the file is saved — the Android share sheet has no built-in save target like the iOS "Save to Files" option.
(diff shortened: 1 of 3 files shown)


alexfreska:
* **#746** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/746?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/746?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

telestrial (approved):

PR #749, opened 2026-06-16
fix(mobile): stop the custom indexer field from clearing and lock it to https

alexfreska:
Android users reported the custom indexer URL field wiping itself on every keystroke (a focus effect keyed on an unstable callback re-ran and cleared the input as they typed), making it impossible to enter a custom indexer at all. This fixes the field so typing sticks, and reframes the input around a locked `https://` prefix shown in the UI: users now enter just the host, any scheme they paste is stripped, and the connection always resolves to an `https://` URL — matching indexd's assumption that indexers are reached over https.

### DIFF
--- a/.changeset/indexer-input-https.md
+++ b/.changeset/indexer-input-https.md
@@ -1,0 +1,5 @@
+---
+'@siastorage/mobile': patch
+---
+
+Fix the custom indexer field clearing itself as you typed, and lock the protocol to https. The URL field now shows a fixed `https://` prefix so you only enter the host, and any pasted protocol is stripped.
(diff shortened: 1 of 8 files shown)


alexfreska:
* **#749** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/749?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/749?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

telestrial (approved):

PR #751, opened 2026-06-16
fix(mobile): scroll PDFs vertically on Android in the file viewer

alexfreska:
On Android, PDFs opened in the file viewer couldn't be scrolled because the surrounding carousel's gesture handler swallowed the touches at the root before the native PDF view ever saw them — iOS was unaffected. The fix wraps the native PDF view in a `Gesture.Native()` detector on Android only, letting it claim vertical scroll touches alongside the carousel's horizontal swipe-to-navigate gesture.

### DIFF
--- a/.changeset/pdf-scroll-android.md
+++ b/.changeset/pdf-scroll-android.md
@@ -1,0 +1,5 @@
+---
+'@siastorage/mobile': patch
+---
+
+Fix PDF scrolling on Android. PDF documents can now be scrolled vertically in the file viewer; previously the swipe-to-navigate gesture prevented scrolling.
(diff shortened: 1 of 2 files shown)


alexfreska:
* **#751** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/751?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/751?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

telestrial (approved):
Good find. Might be good to look at other gestures and make sure they either a) still work with this change b) don't work and we're fine with that.

Pitch to zoom, moving inside that pinch to zoom, swiping to new items, swiping down to exit. That sort of thing. Gesture capturing in the FileViewer has always been precarious, and so these are definitely worth looking at, if they haven't been.

alexfreska:
@telestrial did some manual testing and this quick fix doesn't work, I'm going to close this and file an issue so we can hit it with other viewer work.

PR #750, opened 2026-06-16
fix(mobile,core): exclude thumbnails from status-line file counts

alexfreska:
The status line counted auto-generated thumbnails as if they were files, so backing up a single photo could read "Uploading 3 files" and leave users confused about what was actually happening. It now counts only real files in the "Encrypting/Uploading/Importing N files" message; thumbnails still upload, but when they're all that's left the status keeps showing activity without a misleading count.

PR #753, opened 2026-06-17
refactor(core,mobile): replace sync-up cursor with per-object needsSyncUp dirty flag

alexfreska:
Sync-up tracked progress with a single high-water-mark cursor, so after a large sync-down it re-walked the whole library and re-pushed objects that already matched the server — and worse, an edit landing as the cursor advanced could slip past the mark and never sync at all. It now flags each object dirty on any local change and clears that flag only once the push succeeds, so sync-up pushes only what actually changed and an edit that races a sync is never lost.

Overview
- Schema — new migration 0002_object_needs_sync_up adds objects.needsSyncUp with a partial index and backfills existing rows.
- Object ops — new dirty-flag surface in localObjects: flag on mutation, clear after a successful push, plus the batch query/count that drives sync-up.
- Mutation sites — every local edit (rename, tag add/remove/favorite/rename/delete, trash/restore, tombstone, file/directory moves) now flags the affected file's objects in the same transaction.
- Sync-up — rewritten to walk only dirty objects, replacing the old cursor and its persisted storage; tombstoned files delete the remote object and drop the local row.
- Sync-down — refreshes object metadata while preserving a locally-newer object's pending flag, clearing only the objects the server won.
- Mobile — the "resync metadata" control and data reset now drive the per-object flag ops.
- Tests — rewrote the core op suites (files/trash/tags/directories) and mobile manager suites (syncUpMetadata/syncDownEvents/uploader) to assert the per-object flag, plus a new deleteTag-flags-objects case.


alexfreska:
* **#753** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/753?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/753?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/core/src/app/service.ts
```diff
@@ -402,8 +417,15 @@ export interface AppService {
...
-    /** Creates or updates multiple local objects. */
-    upsertMany(objects: LocalObject[], opts?: { skipInvalidation?: boolean }): Promise<void>
+    /**
+     * Creates or updates multiple local objects. `setNeedsSyncUp` controls the
+     * dirty flag: upload passes 1; sync-down omits it to preserve a pending flag
+     * while refreshing the object metadata.
+     */
+    upsertMany(
+      objects: LocalObject[],
+      opts?: { skipInvalidation?: boolean; setNeedsSyncUp?: 0 | 1 },
```
telestrial: Could be good to unify this with the boolean on line 225, just for consistency's sake, unless there's a good reason it should be different.

packages/core/src/db/operations/files.ts
```diff
@@ -546,10 +568,6 @@ export async function updateFile(
     assignments.updatedAt = Date.now()
   }
 
-  if (!Object.keys(assignments).length) {
```
telestrial, line 549: Is this otherwise caught now or no longer relevant?

packages/core/src/services/syncUpMetadata.ts
```diff
@@ -121,157 +122,142 @@ export async function syncUpMetadataBatch(
...
+          ctx,
+        )
+        if (result === null) return
+        await app.localObjects.clearIfUnchanged(row.objectId, indexerURL, local.updatedAt)
+      }),
+    ),
   )
+  // Progress for the UI: objects attempted this pass, not unique rows cleared. A
+  // failed object stays flagged and is re-counted next pass, so this can exceed
+  // syncUpTotal under sustained retries.
```
telestrial, line 244: Is this a source of the "more operations than a user expects" sort of UI concern we've heard or is this concern elsewhere squashed/irrelevant?

telestrial (approved):

packages/core/src/app/service.ts
alexfreska: Good call, made it a boolean to match `tombstone`.

packages/core/src/db/operations/files.ts
alexfreska: Yeah, intentional. `sql.update` no-ops on an empty set anyway, and returning early would skip the new object flag, which needs to run on every mutation.

packages/core/src/services/syncUpMetadata.ts
alexfreska: Nope, different thing. That was thumbnails getting counted as files in the status line ("Uploading 3 files" for one photo), fixed in #750.

PR #760, opened 2026-06-25
feat(core): WIP

alexfreska:
- WIP: going to adjust this PR.

PR #759, opened 2026-06-25
chore: rust workspace

alexfreska:
The empty Cargo workspace the rest of the stack builds on: the manifest with its one sia-storage-core member, shared package settings and lints, the pinned toolchain, a gitignore, a README, and the lockfile. No dependencies yet.

CI runs fmt, clippy, and tests as three parallel jobs on any change under crates/. A .prettierignore entry keeps the JS formatter off the Rust tree.

PR #767, opened 2026-06-29
feat(core): database engine

alexfreska:
The SQLite engine. Ops are plain sync functions that take a `Connection` and write their statements as literal SQL, and a unit of work gets wrapped in `db.transaction`, where everything inside is one transaction. Async only lives at that boundary: the closure runs on a blocking thread so the synchronous rusqlite calls don't stall the runtime. It leans on rusqlite directly and keeps the transaction boundary explicit, which is the idiomatic shape and leaves fewer ways to misuse it, no ambient state and no surprise nesting.

The everyday shape, a composed write:

```rust
db.transaction(|tx| {
    update_file(tx, &update)?;
    insert_object(tx, &object)?;
    Ok(())
}).await?;
```

Commits if the closure returns `Ok`, rolls back on any error or panic (it's rusqlite's `Transaction`). Reads use the same method, so a multi-statement read sees one consistent snapshot:

```rust
let files = db.transaction(|tx| query_files(tx, &opts)).await?;
```

What it adds:

- `Db` with async `open` / `open_in_memory` (WAL, synchronous=NORMAL, wal_autocheckpoint=500, busy_timeout=5000, foreign keys on, the same pragmas the TS app uses on the shared file), one writer connection behind a `tokio::sync::Mutex`: concurrent transactions queue as parked futures, and only the lock holder occupies a blocking-pool thread.
- `transaction(|tx| ...)`: acquires the connection, then runs one transaction on a blocking thread, commit on `Ok`, rollback on `Err` or panic. Reads and writes both go through it, so a multi-statement read sees a consistent snapshot. The body is sync, so a dropped future can't strand a spawned transaction half-open.
- `interrupt()`: aborts the in-flight statement from any thread for suspend, with the rollback and WAL-lock release in the same unwind.
- `sql`: three small utilities, `escape_like_pattern` for `LIKE` patterns, `id_array` to bind an id list to `rarray(?)` for `IN` membership, and `placeholders` for the rare `IN (...)` whose params mix with other bind values. Everything else is written as literal SQL at the call site.

The migration runner lives in the domain PR with the schema it applies; opening the database runs the migrations.

### DIFF
--- a/crates/sia-storage-core/Cargo.toml
+++ b/crates/sia-storage-core/Cargo.toml
@@ -6,6 +6,11 @@ rust-version.workspace = true
 license.workspace = true
 publish.workspace = true
 
+[dependencies]
+rusqlite.workspace = true
+thiserror.workspace = true
+tokio.workspace = true
+
 [lints]
 workspace = true
 
--- a/crates/sia-storage-core/src/db.rs
+++ b/crates/sia-storage-core/src/db.rs
@@ -1,0 +1,29 @@
+//! The SQLite layer: the [`Db`](database::Db) handle and a few SQL utilities. Operations take a
+//! `&Connection` (reads and writes) or `&mut Transaction` (to open a savepoint) and run inside
+//! the transaction that `Db::transaction` opens for them.
+
+pub mod database;
+pub mod sql;
+
+/// A database-layer error: a SQLite failure, a blocking-thread join failure, or an ad-hoc message.
+#[derive(Debug, thiserror::Error)]
+pub enum DbError {
+    #[error(transparent)]
+    Sqlite(#[from] rusqlite::Error),
+    #[error(transparent)]
+    Join(#[from] tokio::task::JoinError),
+    #[error("{0}")]
+    Message(String),
+}
+
+impl From<&str> for DbError {
+    fn from(s: &str) -> Self {
+        DbError::Message(s.to_string())
+    }
+}
+
+impl From<String> for DbError {
+    fn from(s: String) -> Self {
+        DbError::Message(s)
+    }
+}
--- a/crates/sia-storage-core/src/db/database.rs
+++ b/crates/sia-storage-core/src/db/database.rs
@@ -1,0 +1,177 @@
+use std::sync::Arc;
+
+use rusqlite::{Connection, InterruptHandle, Transaction};
+use tokio::sync::Mutex;
+
+use crate::db::DbError;
+
+/// The SQLite database: one connection plus a thread-safe interrupt handle.
+///
+/// [`transaction`](Db::transaction) runs its closure on a blocking thread so the synchronous
+/// rusqlite calls never block the async runtime. The connection is behind an async mutex, so
+/// concurrent callers wait as parked futures; only the holder occupies a blocking thread.
+pub struct Db {
+    conn: Arc<Mutex<Connection>>,
+    interrupt: InterruptHandle,
+}
+
+impl Db {
+    /// Open the database at `path`, creating if absent. WAL mode, foreign keys on.
+    pub async fn open(path: &str) -> Result<Self, DbError> {
+        let path = path.to_string();
+        tokio::task::spawn_blocking(move || Self::from_conn(Connection::open(path)?)).await?
+    }
+
+    /// Open a private in-memory database with the same pragmas. For tests and ephemeral use.
+    pub async fn open_in_memory() -> Result<Self, DbError> {
+        tokio::task::spawn_blocking(|| Self::from_conn(Connection::open_in_memory()?)).await?
+    }
+
+    fn from_conn(conn: Connection) -> Result<Self, DbError> {
+        conn.pragma_update(None, "journal_mode", "WAL")?;
+        // Drops the per-commit fsync (the slowest part of a write); the cost is that a
+        // full-OS crash can lose the newest commits, though never corrupt the database.
+        conn.pragma_update(None, "synchronous", "NORMAL")?;
+        // Checkpoint at ~2MB, half the default: bounds the on-disk WAL and keeps each
+        // checkpoint's fsync short.
+        conn.pragma_update(None, "wal_autocheckpoint", 500)?;
+        // Multiple processes share this file: wait up to 5s for a lock instead of
+        // failing immediately with SQLITE_BUSY.
+        conn.pragma_update(None, "busy_timeout", 5000)?;
+        conn.pragma_update(None, "foreign_keys", "ON")?;
+        // Registers rarray() so `WHERE col IN rarray(?)` can bind an id list as one parameter.
+        rusqlite::vtab::array::load_module(&conn)?;
+        let interrupt = conn.get_interrupt_handle();
+        Ok(Self {
+            conn: Arc::new(Mutex::new(conn)),
+            interrupt,
+        })
+    }
+
+    /// Abort the statement currently running on the connection, callable from any thread. The
+    /// interrupted statement errors, and the open transaction's `Drop` rolls back and releases the
+    /// WAL write lock in the same unwind. A no-op when the connection is idle.
+    pub fn interrupt(&self) {
+        self.interrupt.interrupt();
+    }
+
+    /// Runs a closure in one transaction: commits on `Ok`, rolls back on `Err` or panic. Reads and
+    /// writes both go through here, so a read-only closure sees a consistent snapshot across its
+    /// statements. A caller cancelled while still waiting for the connection never starts; once
+    /// the closure is spawned, a dropped future still finishes the transaction on its blocking
+    /// thread rather than leaving it open.
+    pub async fn transaction<T, F>(&self, f: F) -> Result<T, DbError>
+    where
+        F: FnOnce(&mut Transaction) -> Result<T, DbError> + Send + 'static,
+        T: Send + 'static,
+    {
+        // Acquired before spawn_blocking: waiters park as futures, and only the
+        // lock holder occupies a blocking-pool thread.
+        let mut guard = self.conn.clone().lock_owned().await;
+        tokio::task::spawn_blocking(move || {
+            let mut tx = guard.transaction()?;
+            let value = f(&mut tx)?;
+            tx.commit()?;
+            Ok(value)
+        })
+        .await?
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    async fn seeded() -> Db {
+        let db = Db::open_in_memory().await.unwrap();
+        db.transaction(|tx| {
+            tx.execute_batch("CREATE TABLE t (id TEXT)")?;
+            Ok(())
+        })
+        .await
+        .unwrap();
+        db
+    }
+
+    async fn ids(db: &Db) -> Vec<String> {
+        db.transaction(|tx| {
+            let mut stmt = tx.prepare("SELECT id FROM t ORDER BY id")?;
+            let rows = stmt.query_map([], |r| r.get(0))?;
+            Ok(rows.collect::<rusqlite::Result<Vec<String>>>()?)
+        })
+        .await
+        .unwrap()
+    }
+
+    #[tokio::test]
+    async fn commits_on_ok() {
+        let db = seeded().await;
+        db.transaction(|tx| {
+            tx.execute("INSERT INTO t (id) VALUES ('a')", [])?;
+            Ok(())
+        })
+        .await
+        .unwrap();
+        assert_eq!(ids(&db).await, ["a"]);
+    }
+
+    #[tokio::test]
+    async fn rolls_back_the_whole_unit_on_error() {
+        let db = seeded().await;
+        let r = db
+            .transaction(|tx| {
+                tx.execute("INSERT INTO t (id) VALUES ('a')", [])?;
+                Err::<(), DbError>("boom".into())
+            })
+            .await;
+        assert!(r.is_err());
+        assert!(
+            ids(&db).await.is_empty(),
+            "the failed unit left nothing behind"
+        );
+    }
+
+    #[tokio::test]
+    async fn savepoint_rolls_back_only_its_substep() {
+        let db = seeded().await;
+        db.transaction(|tx| {
+            tx.execute("INSERT INTO t (id) VALUES ('a')", [])?;
+            let sp = tx.savepoint()?;
+            sp.execute("INSERT INTO t (id) VALUES ('b')", [])?;
+            drop(sp); // no commit -> ROLLBACK TO, undoing 'b' but not the outer 'a'
+            Ok(())
+        })
+        .await
+        .unwrap();
+        assert_eq!(ids(&db).await, ["a"]);
+    }
+
+    // Concurrent transactions serialize on the connection; both land.
+    #[tokio::test]
+    async fn concurrent_transactions_serialize() {
+        let db = std::sync::Arc::new(seeded().await);
+        let a = {
+            let db = db.clone();
+            tokio::spawn(async move {
+                db.transaction(|tx| {
+                    tx.execute("INSERT INTO t (id) VALUES ('a')", [])?;
+                    Ok(())
+                })
+                .await
+            })
+        };
+        let b = {
+            let db = db.clone();
+            tokio::spawn(async move {
+                db.transaction(|tx| {
+                    tx.execute("INSERT INTO t (id) VALUES ('b')", [])?;
+                    Ok(())
+                })
+                .await
+            })
+        };
+        a.await.unwrap().unwrap();
+        b.await.unwrap().unwrap();
+        assert_eq!(ids(&db).await, ["a", "b"]);
+    }
+}
--- a/crates/sia-storage-core/src/db/sql.rs
+++ b/crates/sia-storage-core/src/db/sql.rs
@@ -1,0 +1,64 @@
+//! Small SQL utilities. Statements are written out at their call sites; these only cover what a
+//! literal can't say: binding an id list to `rarray`, escaping LIKE patterns, and placeholder
+//! lists for the rare `IN (...)` that can't use `rarray`.
+
+use std::rc::Rc;
+
+use rusqlite::types::Value;
+
+/// `?, ?, ?` for `n` slots, for an `IN (...)` list whose params mix with other bind values.
+pub fn placeholders(n: usize) -> String {
+    std::iter::repeat_n("?", n).collect::<Vec<_>>().join(", ")
+}
+
+/// The given ids as one array value bindable to `rarray(?)`, for `WHERE col IN rarray(?)`. The
+/// connection must be opened through [`Db`](super::database::Db), which registers the `rarray` module.
+pub fn id_array(ids: &[String]) -> Rc<Vec<Value>> {
+    Rc::new(ids.iter().map(|s| Value::Text(s.clone())).collect())
+}
+
+/// Escape `%`, `_`, and `\` for a `LIKE ? ESCAPE '\'` pattern.
+pub fn escape_like_pattern(s: &str) -> String {
+    let mut out = String::with_capacity(s.len());
+    for c in s.chars() {
+        if matches!(c, '%' | '_' | '\\') {
+            out.push('\\');
+        }
+        out.push(c);
+    }
+    out
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn escape_like_pattern_escapes_wildcards_and_backslash() {
+        assert_eq!(escape_like_pattern("normal"), "normal");
+        assert_eq!(escape_like_pattern("50%_off"), "50\\%\\_off");
+        assert_eq!(escape_like_pattern("a\\b"), "a\\\\b");
+    }
+
+    #[test]
+    fn id_array_filters_an_in_clause() {
+        let c = rusqlite::Connection::open_in_memory().unwrap();
+        rusqlite::vtab::array::load_module(&c).unwrap();
+        c.execute_batch("CREATE TABLE t (id TEXT PRIMARY KEY)")
+            .unwrap();
+        for id in ["a", "b", "c"] {
+            c.execute("INSERT INTO t (id) VALUES (?)", rusqlite::params![id])
+                .unwrap();
+        }
+        let ids = vec!["a".to_string(), "c".to_string()];
+        let mut stmt = c
+            .prepare("SELECT id FROM t WHERE id IN rarray(?) ORDER BY id")
+            .unwrap();
+        let got: Vec<String> = stmt
+            .query_map([id_array(&ids)], |r| r.get(0))
+            .unwrap()
+            .map(Result::unwrap)
+            .collect();
+        assert_eq!(got, ["a", "c"]);
+    }
+}
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -3,5 +3,243 @@
 version = 4
 
 [[package]]
+name = "ahash"
+version = "0.8.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5a15f179cd60c4584b8a8c596927aadc462e27f2ca70c04e0071964a73ba7a75"
+dependencies = [
+ "cfg-if",
+ "once_cell",
+ "version_check",
+ "zerocopy",
+]
+
+[[package]]
+name = "bitflags"
+version = "2.13.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b4388bee8683e3d04af747c73422af53102d2bd24d9eadb6cbc100baef4b43f8"
+
+[[package]]
+name = "cc"
+version = "1.2.65"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e228eec9be7c17ccb640b59b36a5cd805ea2a564a4c5e162c2f659fea30d3b96"
+dependencies = [
+ "find-msvc-tools",
+ "shlex",
+]
+
+[[package]]
+name = "cfg-if"
+version = "1.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
+
+[[package]]
+name = "fallible-iterator"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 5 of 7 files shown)


alexfreska:
* **#770** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/770?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#768** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/768?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#767** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/767?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/767?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#759** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/759?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#760](https://github.com/SiaFoundation/sia-storage-app/pull/760) <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/760?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

n8mgr:
> a nested call joins the open transaction instead of opening a second, tracked by a task-local boolean, so nested operations share one atomic unit.

This raises a question of why are nested txns being started in the first place? The persistence design should handle that. Arbitrarily combining them seems like an atomicity bug waiting to happen.

alexfreska:
@n8mgr it allows for:  
```  
fn update_file_with_local_object(...) {
  with_tx(db, async {
    update_file(db, ...).await?;   // opens its own with_tx, often called standalone, here it joins the outer tx
    insert_object(db, ...).await
  })
}  
```  
but agree that this implementation is not ideal. I'm drafting an update that drops the adapter abstraction (swapping implementations was important in typescript) and uses rusqlite directly which should remove the task-local bit and simplify much of the PR while allowing similar composition.

alexfreska:
@n8mgr ok, updated it. Removed that pattern and dropped the extra abstraction that was carried over from TS, if we ever need it we can add it back in a more rust-native way. Should be much more idiomatic and explicit now, new DX shown in the description.

crates/sia-storage-core/src/db/mod.rs
```diff
@@ -0,0 +1,12 @@
...
+//! inside the transaction `Db::write` opens for them.
+
+pub mod database;
+pub mod runner;
+pub mod sql;
+pub mod types;
+
+/// A database error. Boxed so any source (rusqlite, a blocking-join failure, a typed op error)
+/// flows through `?` without a shared error enum.
+pub type DbError = Box<dyn std::error::Error + Send + Sync>;
```
n8mgr: If the error enums can't compose cleanly, that's likely a design issue to fix rather than papering over with a generic `Box`. We don't really need a generic db abstraction module under our database to where generic errors are necessary. We control the callers. Use `thiserror` and an error enum.

crates/sia-storage-core/src/db/sql.rs
```diff
@@ -0,0 +1,506 @@
...
+
+use rusqlite::types::{ToSql, ToSqlOutput, Value, ValueRef};
+use rusqlite::{Connection, params_from_iter};
+
+use crate::db::DbError;
+
+/// A bindable value. `Bool` binds as integer 0/1 (SQLite has no native boolean). `Null` is emitted
+/// as a literal `NULL` in the SQL rather than bound, so it never consumes a parameter slot.
+#[derive(Debug, Clone)]
+pub enum SqlValue {
```
n8mgr: Is this not just `rusqlite::types::Value`?

crates/sia-storage-core/src/db/sql.rs
```diff
@@ -0,0 +1,506 @@
...
+) -> (Vec<&'static str>, Vec<&'static str>, Vec<&'a SqlValue>) {
+    let mut columns = Vec::with_capacity(row.len());
+    let mut slots = Vec::with_capacity(row.len());
+    let mut params = Vec::with_capacity(row.len());
+    for (col, value) in row {
+        columns.push(*col);
+        if value.is_null() {
+            slots.push("NULL");
+        } else {
+            slots.push("?");
```
n8mgr: `?` should work with `NULL` fine? Is this just an array length optimization? Seems premature.

crates/sia-storage-core/src/db/sql.rs
```diff
@@ -0,0 +1,506 @@
...
+/// Chunk-size that keeps a multi-row statement under `MAX_VARIABLES` bound parameters.
+fn rows_per_chunk(columns: usize) -> usize {
+    (MAX_VARIABLES / columns.max(1)).max(1)
+}
+
+pub fn insert_many(
+    conn: &Connection,
+    table: &str,
+    rows: Vec<Vec<(&'static str, SqlValue)>>,
+    options: InsertOptions,
```
n8mgr: For the most part a prepared statement + loop will get you 80% of the performance with no extra machinery. I'd only go this route if the extra boost is quantifiable instead of preparing for it ahead of time.

crates/sia-storage-core/src/db/sql.rs
```diff
@@ -0,0 +1,506 @@
...
+
+/// The outcome of a write statement.
+#[derive(Debug, Clone, Copy)]
+pub struct SqlRunResult {
+    pub changes: i64,
+    pub last_insert_row_id: i64,
+}
+
+/// `?, ?, ?` for `n` slots, for building an `IN (...)` list.
+pub fn placeholders(n: usize) -> String {
```
n8mgr, line 10: There seems to be an `array` feature in rusqlite that would remove this. This is fine, but it is worth considering that sqlite does have a limit to the number of placeholders if you start doing anything crazy bulky.

n8mgr (changes requested):

crates/sia-storage-core/src/db/mod.rs
alexfreska: Ok yeah I see that, switched it to a thiserror enum.

crates/sia-storage-core/src/db/sql.rs
alexfreska: This was carried over from an android bug where expo-sqlite errored on bound nulls. Doesn't apply with rusqlite, removed.

crates/sia-storage-core/src/db/sql.rs
alexfreska: pretty much yeah, it's just Value plus a Bool variant (sqlite has no bool) and a couple Option conversions the row builders use a lot.

with:

```
("isThumb",   SqlValue::Bool(e.is_thumb)),
("trashedAt", e.trashed_at.into()),
```

without:

```
("isThumb",   Value::Integer(e.is_thumb as i64)),
("trashedAt", e.trashed_at.map_or(Value::Null, Value::Integer)),
```

n8mgr: SQLite automatically coerces `true` and `false` internally. I don't think this is necessary.

alexfreska: Ok let me try that

n8mgr: It looks like ToSql and FromSql are implemented for Rust's bool type already https://docs.rs/rusqlite/latest/rusqlite/types/trait.ToSql.html. Unless I'm mistaken, you should just be able to pass `true` and `false` into the query args directly.

crates/sia-storage-core/src/db/sql.rs
alexfreska: I added the array feature and moved the IN-lists onto it, placeholder still necessary for VALUES but I moved the bulk to prepared statements.

crates/sia-storage-core/src/db/sql.rs
alexfreska: Interesting I didn't realize that, switched it out for the loop.

crates/sia-storage-core/src/db/sql.rs
alexfreska: Ok yeah confirmed it does everything we need, I'll remove this whole extra layer and use rusqlite directly.

crates/sia-storage-core/src/db/database.rs
```diff
@@ -0,0 +1,144 @@
...
+            let mut tx = guard.transaction()?;
+            let value = f(&mut tx)?;
+            tx.commit()?;
+            Ok(value)
+        })
+        .await?
+    }
+
+    /// Run a read closure off the async runtime. Opens no transaction.
+    pub async fn read<T, F>(&self, f: F) -> Result<T, DbError>
```
n8mgr: If `read` ever executes more than one query, it should be in a transaction. Is there any reason to differentiate read and write?

n8mgr (changes requested):
This PR is very generic and probably a bit unnecessarily.

crates/sia-storage-core/src/db/database.rs
alexfreska: Collapsed read and write into just transaction

crates/sia-storage-core/src/db/runner.rs
```diff
@@ -0,0 +1,97 @@
...
+    for m in migrations {
+        if applied.contains(&m.id) {
+            continue;
+        }
+        info!(target: "db", id = %m.id, description = %m.description, "applying_migration");
+        let tx = conn.transaction()?;
+        (m.up)(&tx)?;
+        tx.execute(
+            "INSERT INTO migrations (id, appliedAt) VALUES (?, ?)",
+            rusqlite::params![m.id, chrono::Utc::now().timestamp_millis()],
```
Alrighttt: You could use `std::time::SystemTime::now().duration_since(UNIX_EPOCH)` here and remove the chrono crate.

crates/sia-storage-core/src/db/sql.rs
```diff
@@ -0,0 +1,523 @@
...
+        placeholders(row.len()),
+    );
+    let changes = conn.execute(&sql, params_from_iter(row_params(&row)))?;
+    Ok(SqlRunResult {
+        changes: changes as i64,
+        last_insert_row_id: conn.last_insert_rowid(),
+    })
+}
+
+pub fn insert_many(
```
Alrighttt: Could you add a comment here stating that `table` must never come from unsanitized user input? There are no cases right now where this happens, but using this function incorrectly could allow for SQL injections via the table name.

crates/sia-storage-core/src/db/sql.rs
```diff
@@ -0,0 +1,523 @@
...
+        placeholders(first.len()),
+    );
+    let mut stmt = conn.prepare(&sql)?;
+    for row in &rows {
+        stmt.execute(params_from_iter(row_params(row)))?;
+    }
+    Ok(())
+}
+
+pub fn upsert_many(
```
Alrighttt: Could you add a comment here stating that `table` must never come from unsanitized user input. There are no cases right now where this happens, but using this function incorrectly could allow for SQL injections via the table name.

crates/sia-storage-core/src/db/sql.rs
```diff
@@ -0,0 +1,523 @@
...
+fn column_list(row: &[(&'static str, Value)]) -> String {
+    row.iter().map(|(c, _)| *c).collect::<Vec<_>>().join(", ")
+}
+
+/// The bound values of a `(column, value)` row, in order. NULL binds like any value.
+fn row_params<'a>(row: &'a [(&'static str, Value)]) -> Vec<&'a Value> {
+    row.iter().map(|(_, v)| v).collect()
+}
+
+pub fn insert(
```
Alrighttt: Could you add a comment here stating that `table` must never come from unsanitized user input. There are no cases right now where this happens, but using this function incorrectly could allow for SQL injections via the table name.

crates/sia-storage-core/src/db/sql.rs
```diff
@@ -0,0 +1,523 @@
...
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct UpsertOptions {
+    pub conflict_column: &'static str,
+    pub update_columns: Vec<&'static str>,
+}
+
+fn insert_verb(conflict: Option<InsertConflictClause>) -> String {
```
n8mgr: Why have this extra machinery? Just write the query?

crates/sia-storage-core/src/db/sql.rs
n8mgr: Why? Just write the query?

crates/sia-storage-core/src/db/sql.rs
```diff
@@ -0,0 +1,523 @@
...
+pub fn insert_many(
+    conn: &Connection,
+    table: &str,
+    rows: Vec<Vec<(&'static str, Value)>>,
+    conflict: Option<InsertConflictClause>,
+) -> Result<(), DbError> {
+    let Some(first) = rows.first() else {
+        return Ok(());
+    };
+    let sql = format!(
```
n8mgr: Why? just write the query?

n8mgr (changes requested):
So I started noticing that 90% of my review is probably going to end up requesting the removal of all the helper boilerplate. My general stance is that writing the query is preferred. Note how our SQLite stores in Go work -- we don't add a lot of extra machinery on top to save writing `INSERT` a few times. Keeping the store interface simple and isolated makes things a lot easier to debug and optimize in the long run. We don't need a pseudo-ORM or generic helpers to do some basic db work. More indirection is generally bad when dealing with databases. 

Keep the helpers small and isolated and write out queries so they are maximally grep-able. Once you start defining the public API of the store, it should all just be single method calls at the call site instead of something unnecessarily composable.

crates/sia-storage-core/src/db/sql.rs
```diff
@@ -0,0 +1,523 @@
...
+            parts.push(format!("{col} IS NULL"));
+        } else {
+            parts.push(format!("{col} = ?"));
+            params.push(value);
+        }
+    }
+    (format!(" WHERE {}", parts.join(" AND ")), params)
+}
+
+pub fn update(
```
Alrighttt: Could you add a comment here stating that table must never come from unsanitized user input. There are no cases right now where this happens, but using this function incorrectly could allow for SQL injections via the table name.

crates/sia-storage-core/src/db/sql.rs
```diff
@@ -0,0 +1,523 @@
...
+    let sql = format!("UPDATE {table} SET {assignments}{where_clause}");
+    let changes = conn.execute(&sql, params_from_iter(params.iter()))?;
+    Ok(SqlRunResult {
+        changes: changes as i64,
+        // UPDATE has no insert rowid; conn.last_insert_rowid() would be a stale prior value.
+        last_insert_row_id: 0,
+    })
+}
+
+pub fn del(
```
Alrighttt: Could you add a comment here stating that table must never come from unsanitized user input. There are no cases right now where this happens, but using this function incorrectly could allow for SQL injections via the table name.

crates/sia-storage-core/src/db/sql.rs
```diff
@@ -0,0 +1,523 @@
...
+        conflict_action,
+    );
+    let mut stmt = conn.prepare(&sql)?;
+    for row in &rows {
+        stmt.execute(params_from_iter(row_params(row)))?;
+    }
+    Ok(())
+}
+
+pub fn run(conn: &Connection, sql: &str, params: Vec<Value>) -> Result<SqlRunResult, DbError> {
```
Alrighttt: Could we add a comment here stating that this `sql` must be sanitized?

crates/sia-storage-core/src/db/runner.rs
```diff
@@ -0,0 +1,97 @@
+use rusqlite::Connection;
+use tracing::{debug, info};
+
+use crate::db::DbError;
+use crate::db::types::Migration;
+
+/// Apply every not-yet-applied migration in order, each in its own transaction, recording its id in
+/// the `migrations` table. Idempotent: already-applied ids are skipped. A failing migration rolls
+/// back and is left unrecorded, so the next run retries it.
+pub fn run_migrations(conn: &mut Connection, migrations: &[Migration]) -> Result<(), DbError> {
```
Alrighttt: You could make this function `pub(super)` and instead implement `Db::migrate`. That way callers never handle a raw Connection, and we can't accidentally run migrations without the WAL / foreign_keys / rarray setup that `Db::from_conn` does. 

Eg,
```rust
// runner.rs — no longer part of the public API
pub(super) fn run_migrations(conn: &mut Connection, migrations: &[Migration]) -> Result<(), DbError> { ... }
```

```rust
// database.rs
use super::runner::run_migrations;

impl Db {
    pub async fn migrate(&self, migrations: Vec<Migration>) -> Result<(), DbError> {
        let conn = self.conn.clone();
        tokio::task::spawn_blocking(move || {
            let mut guard = conn.lock();
            run_migrations(&mut guard, &migrations)
        })
        .await?
    }
}
```

crates/sia-storage-core/src/db.rs
```diff
@@ -0,0 +1,31 @@
+//! The SQLite layer: the [`Db`](database::Db) handle, the sync SQL helpers, and the migration
+//! runner. Operations take a `&Connection` (read) or `&mut Connection` (to open a savepoint) and run
```
Alrighttt: rusqlite is unsupported on `wasm32-unknown-unknown`. If this crate is intended to be used on a WASM target eventually, the ops layer probably shouldn't take `&rusqlite::Connection` directly. We will need a trait abstraction here.

crates/sia-storage-core/src/db/sql.rs
```diff
@@ -0,0 +1,523 @@
...
+    Ok(SqlRunResult {
+        changes: changes as i64,
+        last_insert_row_id: conn.last_insert_rowid(),
+    })
+}
+
+pub fn insert_many(
+    conn: &Connection,
+    table: &str,
+    rows: Vec<Vec<(&'static str, Value)>>,
```
Alrighttt: Column names are only read from the first row here. Every later row's names are silently thrown away, and its values are bound positionally. We could change the arguments to `(columns: &[&'static str], rows: Vec<Vec<Value>>)` so there's only one column list by construction, or otherwise make the function actually use each row's column names.

With the current code...
```rust
let rows = vec![
    vec![("id", Value::Text("a".into())), ("size", Value::Integer(100))],  // path 1
    vec![("size", Value::Integer(200)), ("id", Value::Text("b".into()))],  // path 2
];
insert_many(&conn, "files", rows, None)?;
```

The second row would be silently inserted as `id=200, size='b'` without any error.

Same issue in `upsert_many` below.

alexfreska:
@n8mgr fair, many/most queries do not use the helpers anyway since they are for basic ops, I'll remove the helpers here and update the db ops PRs.

crates/sia-storage-core/src/db/sql.rs
```diff
@@ -0,0 +1,523 @@
...
+        stmt.execute(params_from_iter(row_params(row)))?;
+    }
+    Ok(())
+}
+
+pub fn run(conn: &Connection, sql: &str, params: Vec<Value>) -> Result<SqlRunResult, DbError> {
+    let changes = conn.execute(sql, params_from_iter(params.iter()))?;
+    Ok(SqlRunResult {
+        changes: changes as i64,
+        last_insert_row_id: conn.last_insert_rowid(),
```
Alrighttt: `last_insert_rowid` is only updated by a successful INSERT. UPDATE, DELETE, and DDL leave it untouched. This means that when `run()` executes a statement that is not an INSERT, it reports the rowid of some earlier insert. This is inconsistent with `update()` and `del()`, which both hardcode `0` specifically to avoid returning a stale rowid, so the same UPDATE statement yields a different `SqlRunResult` depending on whether it goes through `update()` or `run()`. The same problem exists in `insert()` with `OrIgnore`, where an ignored conflict returns `changes: 0` but a stale rowid from a previous statement.

crates/sia-storage-core/src/db/sql.rs
```diff
@@ -0,0 +1,523 @@
...
+}
+
+pub fn insert(
+    conn: &Connection,
+    table: &str,
+    row: Vec<(&'static str, Value)>,
+    conflict: Option<InsertConflictClause>,
+) -> Result<SqlRunResult, DbError> {
+    let sql = format!(
+        "{} INTO {} ({}) VALUES ({})",
```
Alrighttt: `insert`, `insert_many`, and `upsert_many` each assemble the same `INSERT INTO {table} ({columns}) VALUES ({placeholders})` string. This could be extracted into a single helper that builds the statement.

crates/sia-storage-core/src/db/database.rs
```diff
@@ -0,0 +1,135 @@
+use std::sync::Arc;
+
+use parking_lot::Mutex;
```
Alrighttt: Consider using `tokio::sync::Mutex` here instead of `parking_lot::Mutex`, and acquiring the lock before entering `spawn_blocking`. Currently, the lock is taken inside the blocking closure, so N concurrent transaction() calls occupy N blocking pool threads while N-1 of them just sit parked waiting for the connection. If the guard is acquired asynchronously first, waiters queue as cheap pending futures and only the transaction actually running holds a blocking thread.

```rust
pub async fn transaction<T, F>(&self, f: F) -> Result<T, DbError>
where
    F: FnOnce(&mut Transaction) -> Result<T, DbError> + Send + 'static,
    T: Send + 'static,
{
    let mut guard = self.conn.clone().lock_owned().await;
    tokio::task::spawn_blocking(move || {
        let mut tx = guard.transaction()?;
        let value = f(&mut tx)?;
        tx.commit()?;
        Ok(value)
    })
    .await?
}
```
`lock_owned()` returns an `OwnedMutexGuard` that is `Send` and derefs mutably, so it can move into the closure and everything else stays the same.

You could remove `parking_lot` as a dependency, and an async mutex is the only variant that could work on a single threaded WASM target. 

One doc tweak: a future dropped while still waiting for the lock now never starts the transaction, so the "dropped future still finishes" comment applies only once the closure has been spawned.

crates/sia-storage-core/src/db/runner.rs
n8mgr: I'd probably go one step further and make migrations const instead of letting callers pass it in. Migrations are a database implementation detail. Exposing them publicly is a leaky abstraction. Potentially consider calling this on open instead of requiring callers to think about it at all.

crates/sia-storage-core/src/db.rs
alexfreska: Had that and removed for now, can discuss more on slack.

crates/sia-storage-core/src/db/database.rs
alexfreska: This is great, applied as suggested, lock_owned before spawn_blocking and parking_lot is gone. Also made open async on the same reasoning, since open now runs migrations (per the runner discussion) and that work belongs on the blocking pool too.

crates/sia-storage-core/src/db/runner.rs
alexfreska: The runner and migration list are now internal to the db module and open runs them, so callers never see a Migration at all. The runner also moved up to the schema PR since it was dead code here without the migrations it applies.

crates/sia-storage-core/src/db/sql.rs
alexfreska: This got removed along with the rest of the helper layer, everything is literal SQL at the call site now.

crates/sia-storage-core/src/db/sql.rs
alexfreska: This got removed along with the rest of the helper layer, everything is literal SQL at the call site now.

crates/sia-storage-core/src/db/sql.rs
alexfreska: This got removed along with the rest of the helper layer, everything is literal SQL at the call site now.

crates/sia-storage-core/src/db/sql.rs
alexfreska: This got removed along with the rest of the helper layer, everything is literal SQL at the call site now.

alexfreska:
@n8mgr removed the helper layer, everything will be literal SQL at the call site now :+1:

crates/sia-storage-core/Cargo.toml
```diff
@@ -6,6 +6,13 @@ rust-version.workspace = true
 license.workspace = true
 publish.workspace = true
 
+[dependencies]
+chrono.workspace = true
```
Alrighttt: chrono is an unused dep

crates/sia-storage-core/Cargo.toml
```diff
@@ -6,6 +6,13 @@ rust-version.workspace = true
...
+rusqlite.workspace = true
+thiserror.workspace = true
+tokio.workspace = true
+tracing.workspace = true
```
Alrighttt: tracing is an unused dep

crates/sia-storage-core/src/db.rs
```diff
@@ -0,0 +1,29 @@
+//! The SQLite layer: the [`Db`](database::Db) handle and a few SQL utilities. Operations take a
+//! `&Connection` (read) or `&mut Connection` (to open a savepoint) and run inside the transaction
```
Alrighttt: Could you update this to say ops "take `&Connection` (reads) or `&mut Transaction` (to open a savepoint)"?

Transaction has no `DerefMut`, so an op taking `&mut Connection` couldn't actually be called inside the closure.

Alrighttt (changes requested):

crates/sia-storage-core/Cargo.toml
alexfreska: Yep removed, this belonged to the runner, which moved up to the schema PR.

crates/sia-storage-core/Cargo.toml
alexfreska: Moved to the schema PR.

crates/sia-storage-core/src/db.rs
alexfreska: Updated, went with "reads and writes" for the first one since writes also go through &Connection, the savepoint case is the only reason anything would take &mut Transaction.

Alrighttt (approved):

n8mgr (approved):

PR #768, opened 2026-06-29
feat(core): domain database

alexfreska:
The core schema and the migration runner, plus the epoch-ms timestamp codec the ops share. Two migrations build the tables and indexes and seed the Favorites tag.

Migrations are an internal detail of the db module: opening the database runs them, and no caller sees the runner or the migration list. Each migration is sync, it runs its statements and the runner wraps them in a transaction:

```rust
fn up(conn: &Connection) -> Result<(), DbError> {
    conn.execute_batch(SCHEMA)?;
    Ok(())
}
```

The runner applies each pending migration in its own transaction and records its id, so already-applied ids are skipped and a failed migration retries on the next open. `open` takes an optional per-migration progress callback (`MigrationProgress`: id, description, index, total) so callers can surface migration progress; the mobile app displays it during init.

The one non-migration piece is the timestamp codec: the DB stores timestamps as epoch-ms `i64`, and `encode_epoch_ms` / `decode_epoch_ms` convert to and from the domain `DateTime`. It lives here because more than one ops file needs it (files and local-objects). The stored types themselves (`LocalObject` and the slab codec) live with their only consumer in the local-objects PR.

What it adds:

- `0001`: the full schema (files / directories / objects / fs / logs / tags / file_tags plus indexes) and the Favorites system tag.
- `0002`: the per-object `needsSyncUp` dirty flag and its partial index.
- The runner: each pending migration in its own transaction, recorded, idempotent; wired into `Db::open` with the optional progress callback.
- The epoch-ms timestamp codec (`encode_epoch_ms` / `decode_epoch_ms`), shared by the ops that read and write timestamps.

### DIFF
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -15,10 +15,31 @@ dependencies = [
 ]
 
 [[package]]
+name = "android_system_properties"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
+dependencies = [
+ "libc",
+]
+
+[[package]]
+name = "autocfg"
+version = "1.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f2032f911046de80f0a198e0901378627c33f59ea0ac00e363d481118bd70a53"
+
+[[package]]
 name = "bitflags"
 version = "2.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b4388bee8683e3d04af747c73422af53102d2bd24d9eadb6cbc100baef4b43f8"
+
+[[package]]
+name = "bumpalo"
+version = "3.20.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "72f5acc6cb2ba439de613abc23857ec3d78374d8ed5ac84e9d11336e87da8649"
 
 [[package]]
 name = "cc"
@@ -35,6 +56,26 @@ name = "cfg-if"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
+
+[[package]]
+name = "chrono"
+version = "0.4.45"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1aa79e62e7697b8e29b513a68abacf485adcd1fe8284a4316c5ae868e6633327"
+dependencies = [
+ "iana-time-zone",
+ "js-sys",
+ "num-traits",
+ "serde",
+ "wasm-bindgen",
+ "windows-link",
+]
+
+[[package]]
+name = "core-foundation-sys"
+version = "0.8.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "773648b94d0e5d620f64f280777445740e61fe701025087ec8b57f45c791888b"
 
 [[package]]
 name = "fallible-iterator"
@@ -55,6 +96,30 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5baebc0774151f905a1a2cc41989300b1e6fbb29aff0ceffa1064fdd3088d582"
 
 [[package]]
+name = "futures-core"
+version = "0.3.32"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7e3450815272ef58cec6d564423f6e755e25379b217b0bc688e295ba24df6b1d"
+
+[[package]]
+name = "futures-task"
+version = "0.3.32"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "037711b3d59c33004d3856fbdc83b99d4ff37a24768fa1be9ce3538a1cde4393"
+
+[[package]]
+name = "futures-util"
+version = "0.3.32"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "389ca41296e6190b48053de0321d02a77f32f8a5d2461dd38762c0593805c6d6"
+dependencies = [
+ "futures-core",
+ "futures-task",
+ "pin-project-lite",
+ "slab",
+]
+
+[[package]]
 name = "hashbrown"
 version = "0.14.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -73,6 +138,48 @@ dependencies = [
 ]
 
 [[package]]
+name = "iana-time-zone"
+version = "0.1.65"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e31bc9ad994ba00e440a8aa5c9ef0ec67d5cb5e5cb0cc7f8b744a35b389cc470"
+dependencies = [
+ "android_system_properties",
+ "core-foundation-sys",
+ "iana-time-zone-haiku",
+ "js-sys",
+ "log",
+ "wasm-bindgen",
+ "windows-core",
+]
+
+[[package]]
+name = "iana-time-zone-haiku"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f31827a206f56af32e590ba56d5d2d085f558508192593743f16b2306495269f"
+dependencies = [
+ "cc",
+]
+
+[[package]]
+name = "js-sys"
+version = "0.3.99"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "142bc4740e452c1e57ade0cbc129f139c9093e354346f0872ef985f4f5cf5f11"
+dependencies = [
+ "cfg-if",
+ "futures-util",
+ "once_cell",
+ "wasm-bindgen",
+]
+
+[[package]]
+name = "libc"
+version = "0.2.186"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "68ab91017fe16c622486840e4c83c9a37afeff978bd239b5293d61ece587de66"
+
+[[package]]
 name = "libsqlite3-sys"
 version = "0.30.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -81,6 +188,21 @@ dependencies = [
  "cc",
  "pkg-config",
  "vcpkg",
+]
+
+[[package]]
+name = "log"
+version = "0.4.33"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0ceec5bc11778974d1bcb055b18002eba7f4b3518b6a0081b3af5f21666da9ad"
+
+[[package]]
+name = "num-traits"
+version = "0.2.19"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "071dfc062690e90b734c0b2273ce72ad0ffa95f0c74596bc250dcfd960262841"
+dependencies = [
+ "autocfg",
 ]
 
 [[package]]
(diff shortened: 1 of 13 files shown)


alexfreska:
* **#770** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/770?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#768** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/768?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/768?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#767** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/767?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#759** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/759?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#760](https://github.com/SiaFoundation/sia-storage-app/pull/760) <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/760?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

crates/sia-storage-core/src/db.rs
n8mgr, line 1: mod.rs is the old structure. This should be `db.rs` in `src/`. Same with the others

crates/sia-storage-core/src/encoding/local_object.rs
```diff
@@ -0,0 +1,160 @@
+use crate::encoding::slabs::{slabs_from_storage_string, slabs_to_storage_string};
+use crate::encoding::timestamp::{decode_epoch_ms, encode_epoch_ms};
+use crate::types::local_object::LocalObject;
+
+/// DB row form. All bytes are hex strings; timestamps are epoch-ms i64s.
+#[derive(Debug, Clone)]
+pub struct LocalObjectRow {
```
n8mgr: Are these wrapper types really necessary? You're effectively keeping two versions in memory when a borrow on the original will likely be fine.

crates/sia-storage-core/src/types/local_object.rs
n8mgr, line 1: Global `types` packages are sad. Types should live close to the code they're needed in.

crates/sia-storage-core/src/types/slabs.rs
```diff
@@ -0,0 +1,23 @@
+use serde::{Deserialize, Serialize};
+
+/// A sector stored on a specific host: its Merkle `root` and the host's `hostKey`.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct PinnedSector {
```
n8mgr: Why define these? The SDK exports a nearly equivalent type. Just to get a String?

crates/sia-storage-core/src/types/slabs.rs
```diff
@@ -0,0 +1,23 @@
...
+    pub host_key: String,
+}
+
+/// One erasure-coded slab of a file: the key that decrypts its data, the minimum
+/// sectors needed to recover it, the sectors holding the shards, and its byte
+/// offset and length within the file. Stored as JSON in the local database;
+/// carried to the indexer inside the object.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Slab {
```
n8mgr: Ditto

crates/sia-storage-core/src/encoding/slabs.rs
```diff
@@ -0,0 +1,160 @@
+use crate::types::slabs::{PinnedSector, Slab};
+use serde::{Deserialize, Serialize};
+
+/// The stored form of a slab: the encryption key is hex, not raw bytes.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct SlabStorage {
```
n8mgr: Is any of this encoding middle-man necessary?

n8mgr (changes requested):

crates/sia-storage-core/src/db.rs
alexfreska: fixed

crates/sia-storage-core/src/types/slabs.rs
alexfreska: Pulled in sia_storage types and removed.

crates/sia-storage-core/src/encoding/slabs.rs
alexfreska: Removed as much as possible, left just what is required for the exact db serialization.

crates/sia-storage-core/src/types/local_object.rs
alexfreska: Removed most of the structs in favor of embedded SealedObject, this last type is imported in many files.

crates/sia-storage-core/src/encoding/local_object.rs
alexfreska: Nuked

crates/Cargo.lock
```diff
@@ -66,264 +227,1640 @@ dependencies = [
...
-name = "num-traits"
-version = "0.2.19"
+name = "darling_macro"
+version = "0.23.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "071dfc062690e90b734c0b2273ce72ad0ffa95f0c74596bc250dcfd960262841"
+checksum = "ac3984ec7bd6cfa798e62b4a642426a5be0e68f9401cfc2a01e3fa9ea2fcdb8d"
 dependencies = [
- "autocfg",
+ "darling_core",
```
alexfreska: sia_storage made this lock file diff huge :sweat_smile:

crates/sia-storage-core/src/types/local_object.rs
alexfreska: Moved the local object and its serializers to the next PR.

crates/sia-storage-core/src/db/migrations/m0001_init_schema.rs
```diff
@@ -0,0 +1,150 @@
...
+CREATE INDEX IF NOT EXISTS idx_files_current_size
+    ON files(size DESC, id DESC)
+    WHERE current = 1 AND kind = 'file' AND trashedAt IS NULL AND deletedAt IS NULL;
+CREATE INDEX IF NOT EXISTS idx_files_current_type
+    ON files(type)
+    WHERE current = 1 AND kind = 'file' AND trashedAt IS NULL AND deletedAt IS NULL;
+CREATE INDEX IF NOT EXISTS idx_files_current_directoryId
+    ON files(directoryId)
+    WHERE current = 1 AND kind = 'file' AND trashedAt IS NULL AND deletedAt IS NULL;
+CREATE INDEX IF NOT EXISTS idx_files_current_name_directoryId
```
peterjan, line 69: We can push the `files.current` invariant down to the DB instead of trusting every write path to maintain it. We can do that by making the existing partial index UNIQUE I think, and a COALESCE on directoryId covers root-level files too.

```
CREATE UNIQUE INDEX IF NOT EXISTS idx_files_current_name_directoryId
    ON files(name, COALESCE(directoryId, ''))
    WHERE current = 1 AND kind = 'file' AND trashedAt IS NULL AND deletedAt IS NULL;
```

Please double check though because it assumes directory ids are never empty. Also it means that you have to clear the old row's current before the new current = 1 row lands.

peterjan (approved):
LGTM 👍

crates/sia-storage-core/src/db/migrations/m0001_init_schema.rs
alexfreska: Nice find, although we can't change this migration since it's already shipped (the rust migrations need to match the mobile/typescript exactly), but a follow-up migration could add it. There's definitely a good fix to the write ordering here regardless: we insert new versions as current=1 (the column default) and recalc after, so today the index would trip on version inserts and trash-restores before the recalc runs. Inserting as current=0 and letting the recalc promote makes the logic consistent with the invariant on its own. On the index itself I want to evaluate a bit more, we tend to lean away from unique constraints in favor of repair logic here (current is derived state the recalc can always fix, and a constraint failure during sync apply would wedge the device instead of self-healing), but I documented all of this in https://github.com/SiaFoundation/sia-storage-app/issues/776

crates/sia-storage-core/src/db/migrations.rs
```diff
@@ -0,0 +1,88 @@
+pub mod m0001_init_schema;
+pub mod m0002_object_needs_sync_up;
+
+use crate::db::types::Migration;
+
+pub fn core_migrations() -> Vec<Migration> {
+    vec![
+        m0001_init_schema::migration_0001_init_schema(),
```
n8mgr: One mod per migration is going to get messy fast. This set design also enforces each migrations to be indempotent with regards to all other migrations, which is a property that rarely lasts. I realize you don't want to change the tables, but consider only executing migrations *after* the last one that has run so that mid-list migrations can't run out of order and leave the database in a weird state (see `renterd` issues). This looks good on paper because merge issues can continue to work, but in pratice it can create issues without a lot of extra care. `renterd` corrupted user databases quite a few times because of a similar design.

crates/sia-storage-core/src/db/migrations.rs
```diff
@@ -0,0 +1,88 @@
+pub mod m0001_init_schema;
+pub mod m0002_object_needs_sync_up;
+
+use crate::db::types::Migration;
+
+pub fn core_migrations() -> Vec<Migration> {
```
n8mgr: Should not be public and should be a const &[Migration].

crates/sia-storage-core/src/db/migrations/m0001_init_schema.rs
```diff
@@ -0,0 +1,150 @@
...
+    conn.execute_batch(SCHEMA)?;
+    let now = chrono::Utc::now().timestamp_millis();
+    conn.execute(
+        "INSERT OR IGNORE INTO tags (id, name, createdAt, usedAt, system) VALUES (?, ?, ?, ?, 1)",
+        params!["sys:favorites", "Favorites", now, now],
+    )?;
+    Ok(())
+}
+
+pub fn migration_0001_init_schema() -> Migration {
```
n8mgr: Why a function? If you want to keep the migration in a module, export `up` as `migration_0001_init_schema` then define the `Migration` type as a const.

crates/sia-storage-core/src/db/types.rs
```diff
@@ -0,0 +1,26 @@
+use rusqlite::Connection;
+
+use crate::db::DbError;
+
+/// A schema migration. `up` runs the migration's statements; the runner wraps them in a transaction.
+pub struct Migration {
+    pub id: String,
+    pub description: String,
+    pub up: fn(&Connection) -> Result<(), DbError>,
+}
```
n8mgr:
```
pub struct Migration<'a> {
    pub id: &'a str,
    pub description: &'a str,
    pub up: fn(&Connection) -> Result<(), DbError>,
}
```

makes this const-able. Should also not be `pub` migrations are an internal database issue, not a consumer issue.

crates/Cargo.toml
```diff
@@ -9,9 +9,15 @@ license = "Apache-2.0"
 publish = false
 
 [workspace.dependencies]
+chrono = { version = "0.4", features = ["serde"] }
+hex = "0.4"
 rusqlite = { version = "0.32", features = ["bundled", "array"] }
+serde = { version = "1", features = ["derive"] }
+serde_json = "1"
+sia_storage = { git = "https://github.com/SiaFoundation/sia-sdk-rs", branch = "master" }
```
n8mgr: Why git? The crate is published.

n8mgr (changes requested):
Lots of unnecessarily `pub`. Make sure we're not leaking database internals unless necessary.

crates/sia-storage-core/src/db/migrations.rs
alexfreska: I was expecting them to only ever run in the specified order, but to make that a hard guarantee I changed it so each migration now declares its parent and the runner only applies one whose parent is the current applied head, so anything out of order errors instead of running.

n8mgr (approved):

alexfreska:
### Merge activity

* **Jul 15, 4:07 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/768).
* **Jul 15, 4:07 PM UTC**: @alexfreska merged this pull request with [Graphite](https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/768).

PR #769, opened 2026-06-30
ci(release): automate TestFlight and Play tester distribution

alexfreska:
Each release currently requries a bunch of manual steps like hand-entering the release notes, assigning the TestFlight tester groups, and clicking through the Play Console to activate the build, this was already tedius but is now 2x because we added a seperate Beta testflight app (so you can have testflight and app store installed at the same time). With these changes CI now handles it all. The testflight notes now link to that version's GitHub release page (with knope managed changelog). We link out because Android has a 500 character limit, the knope changelog almost always exceeds this.

**TestFlight (iOS)**

- Sets "What to Test" to the release link
- Distributes to each app's internal and external groups, with notify on

**Play (Android)**

- Sets "What's new" to the release link
- Ships the internal build as `completed` so it activates without Play Console clicks

**Both patforms**

- Runs automatically for main app and beta, so nothing above is done twice by hand

**Unchanged**

- App Store and Play production release process and notes stay manual

### DIFF
--- a/apps/mobile/fastlane/Fastfile
+++ b/apps/mobile/fastlane/Fastfile
@@ -176,11 +176,18 @@ platform :ios do
     
     app_store_connect_api_key(api_key_params)
     
+    # External tester groups for this build, passed by releaseIos.ts; an empty
+    # list means internal testers only.
+    external_groups = ENV["IOS_EXTERNAL_GROUPS"] ? JSON.parse(ENV["IOS_EXTERNAL_GROUPS"]) : []
+    has_external = !external_groups.empty?
+
     upload_to_testflight(
       ipa: IOS_RELEASE_IPA_PATH,
       skip_waiting_for_build_processing: false, # Wait for processing - ensures build is ready before command completes (takes 10-30 min)
-      distribute_external: false, # Don't auto-distribute externally - we manually manage tester groups in TestFlight
-      notify_external_testers: false # Don't notify testers - we control when to send notifications manually
+      changelog: ENV["IOS_WHATS_NEW"], # "What to Test" - releaseIos.ts sets it to the version's release-notes link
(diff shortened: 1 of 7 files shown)


alexfreska:
* **#769** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/769?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/769?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

apps/mobile/fastlane/Fastfile
```diff
@@ -281,16 +288,23 @@ platform :android do
     if !File.exist?(ANDROID_AAB_PATH)
       UI.user_error!("AAB file not found at #{ANDROID_AAB_PATH}. Run 'bun run release:android:build' first.")
     end
-    
+
+    # releaseAndroid.ts writes this before calling the lane; invoking the lane
+    # on its own (not via the script) leaves it absent, so skip changelogs then.
+    metadata_path = File.expand_path("metadata", __dir__)
```
peterjan: Can you verify android because I think this is one level too high? It should be `metadata/android` I think or you won't get the release notes? Def. no expert but it makes sense to verify because it would defeat the point of the PR entirely

peterjan (changes requested):

apps/mobile/fastlane/Fastfile
alexfreska: oh yeah let me fix that

peterjan (approved):

ChrisSchinnerl (approved):

PR #771, opened 2026-07-02
feat(core): log store db operations

alexfreska:
The on-device log store. Structured rows (timestamp, level, scope, message, optional JSON `data`, plus a stamped `createdAt`) in the `logs` table: written by the app logger, read and exported by the settings log viewer, tailed by the remote forwarder, capped by rotation.

```rust
db.transaction(|tx| insert_log(tx, entry)).await?;
let rows = db.transaction(|tx| query_logs(tx, opts)).await?;
```

What it adds:

Ingestion:

- `insert_log`: insert one log row, for the logger's single-entry writes.
- `insert_many_logs`: batch insert, no-op on empty, for flushing buffered entries.

Settings log viewer:

- `query_logs`: read rows filtered by min-level, scopes, and limit, newest first (`createdAt DESC, id DESC`), for the viewer list and its NDJSON export.
- `query_available_log_scopes`: distinct scopes, sorted, for the scope-filter dropdown.
- `count_logs`: count rows matching the same level/scope filter, for the viewer's total.

Remote forwarder:

- `query_logs_since_id`: up to `limit` rows with `id` above a cursor, ascending, raw `data` string unparsed, for the forwarder's tail cursor.

Maintenance:

- `rotate_logs`: trim to the newest `max_logs` rows (oldest `createdAt` first), returning the delete count, for rotation.
- `delete_all_logs`: clear the table, for the clear-logs action.

`query_logs` and `count_logs` share one filter builder: a min level expands to that level and everything above it (`debug` < `info` < `warn` < `error`, an unknown level falls back to `error` only), combined with a scope `IN` set.


alexfreska:
* **#771** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/771?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/771?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/sia-storage-core/src/db/operations/logs.rs
```diff
@@ -0,0 +1,493 @@
...
+    // An unknown level yields only `["error"]` (the highest), not all four.
+    match LEVEL_ORDER.iter().position(|s| *s == min_level) {
+        Some(i) => &LEVEL_ORDER[i..],
+        None => &LEVEL_ORDER[LEVEL_ORDER.len() - 1..],
+    }
+}
+
+/// Appends one log row. Consumed by the logging bridge (upstack), which stamps
+/// `created_at` immediately before the write.
+pub fn insert_log(conn: &Connection, entry: &LogInsert) -> Result<(), DbError> {
```
peterjan: should call `insert_many_logs`?

crates/sia-storage-core/src/db/operations/logs.rs
```diff
@@ -0,0 +1,493 @@
...
+    pub level: String,
+    pub scope: String,
+    pub message: String,
+    pub data: Option<String>,
+}
+
+const LEVEL_ORDER: &[&str] = &["debug", "info", "warn", "error"];
+
+fn levels_for_filter(min_level: &str) -> &'static [&'static str] {
+    // An unknown level yields only `["error"]` (the highest), not all four.
```
peterjan: instead of mapping to error couldn't we model it as a serde `LogLevel` enum?

crates/sia-storage-core/src/db/operations/logs.rs
```diff
@@ -0,0 +1,493 @@
...
+        "SELECT id, timestamp, level, scope, message, data FROM logs WHERE id > ? ORDER BY id ASC LIMIT ?",
+    )?;
+    let out = stmt
+        .query_map(params![since_id, limit], log_row_with_id_from_db_row)?
+        .collect::<rusqlite::Result<Vec<LogRowWithId>>>()?;
+    Ok(out)
+}
+
+/// Row count under the same filters as [`query_logs`].
+pub fn count_logs(conn: &Connection, opts: LogQueryOpts) -> Result<i64, DbError> {
```
peterjan: `nit` this ignores `limit` on the options, perhaps we could split it into `LogFilterOpts` and `LogQueryOpts`?

alexfreska: updated

crates/sia-storage-core/src/db/operations/logs.rs
alexfreska: updated

crates/sia-storage-core/src/db/operations/logs.rs
alexfreska: There was previous feedback on inlining everything, and its slightly more efficient as a direct exec

crates/sia-storage-core/src/db/operations/logs.rs
```diff
@@ -0,0 +1,619 @@
...
+pub struct LogFilterOpts {
+    pub log_level: Option<LogLevel>,
+    pub log_scopes: Option<Vec<String>>,
+}
+
+#[derive(Debug, Default, Clone, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct LogQueryOpts {
+    pub filter: LogFilterOpts,
+    pub limit: Option<i64>,
```
n8mgr: usize or u32

crates/sia-storage-core/src/db/operations/logs.rs
```diff
@@ -0,0 +1,619 @@
...
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct LogInsert {
+    pub timestamp: String,
+    pub level: LogLevel,
+    pub scope: String,
+    pub message: String,
+    pub data: Option<String>,
+    // IPC payloads omit this field: the facade stamps insertion time right before
```
n8mgr: This comment doesn't seem to be accurate? `insert_log` does not override the timestamp.

alexfreska: That was referring to the app service layer above that consumes these methods, removed the comment and also fixed properly, took createdAt off the API entirely. insert_log stamps it now, so there's nothing for a caller to set or explain here.

crates/sia-storage-core/src/db/operations/logs.rs
alexfreska: okay updated query bounds (limit, offset, page size) to u32 and counts to u64 in this and the rest of the db PRs.

alexfreska:
@n8mgr addressed those two comments.

crates/sia-storage-core/src/db/operations/logs.rs
```diff
@@ -0,0 +1,624 @@
...
+    pub level: LogLevel,
+    pub scope: String,
+    pub message: String,
+    pub data: Option<String>,
+}
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct LogRowWithId {
+    pub id: i64,
```
n8mgr: Probably also u64 here. IDs can't be negative. May just do a quick audit for anything that really shouldn't be a signed integer. Other than that looks good.

alexfreska: This was i64 because that what the sqlite column is, but makes sense for it to be non negative outside of the storage layer. I flipped id + the since_id cursor to u64, all other table ids are strings so no update necessary for those. The remaining i64s are all the timestamps which match what chrono APIs use, so I think we should probably leave those.

alexfreska:
@n8mgr updated the id and since_id to u64.

PR #772, opened 2026-07-02
feat(core): tag db operations

alexfreska:
The tags module: user tags plus the built-in Favorites system tag, the `file_tags` links joining tags to files, tag CRUD, and the sync reconcile paths that keep a file's tags matching indexer metadata. Tagging a file bumps its `updatedAt` and marks the object dirty, so the change re-pushes on the next sync-up. This PR also lands `filter.rs`: `build_record_filter`, the active-record predicate every file, directory, and stats query funnels through, extracted here because the tag counts are its first consumer. Same story for `unique_id` (a UUIDv4 in simple form; ids are opaque and never parsed): tag ids are its first consumer, directories reuse it downstack.

```rust
db.transaction(|tx| add_tag_to_file(tx, &file_id, "Travel")).await?;
let tags = db.transaction(|tx| query_all_tags_with_counts(tx)).await?;
```

What it adds:

System tags
- `system_tags`: the static built-in list (just Favorites); m0001 seeds the row in SQL, so only `ensure_system_tags` reads it.
- `ensure_system_tags`: idempotently seed every system tag with `system = 1`; run at startup and before each sync reconcile.

Tag CRUD (manage-tags)
- `insert_tag`: create a user tag, rejecting a blank or duplicate name.
- `get_or_create_tag`: return the tag by name or create it and bump `usedAt`; the shared insert path for the add and sync ops.
- `rename_tag`: rename a user tag (rejects system tags) and re-push every file carrying it.
- `delete_tag`: drop a user tag and its `file_tags` links (rejects system tags) and re-push the carriers.
- `query_all_tags_with_counts`: list every tag with its live-file count, system tags first.
- `query_tags_by_prefix`: prefix-match tag names for autocomplete, ordered by recency.

Tagging files
- `add_tag_to_file`: create the tag if needed, link it, and bump the file.
- `add_tag_to_files`: same across a set of files in one shot, for multi-select tagging.
- `remove_tag_from_file`: unlink a tag and bump the file.
- `remove_tag_from_files`: batch-unlink a tag from a set of files, the multi-select counterpart to `remove_tag_from_file`.
- `insert_file_tag`: link an existing tag id to a file with no create and no `updatedAt` bump, the low-level link for when the tag id is already known.
- `query_tags_for_file`: a file's tags alphabetically, for file details.
- `query_tag_names_for_file`: a file's tag names or `None` in the indexer encoder's shape; read during the sync-up metadata build so tags reach other devices.

Favorites
- `toggle_favorite`: add or remove the Favorites link and bump the file.
- `query_is_favorite`: report whether a file carries Favorites; also gates `toggle_favorite`.

Sync-down apply
- `sync_tags_from_metadata_single`: reconcile one file's `file_tags` to the canonical names from indexer metadata, keeping system links.
- `sync_tags_from_metadata`: same for a batch, sync-down step 12.

Shared query filter
- `build_record_filter` / `BuildRecordFilterOpts`: build the active-record WHERE clause (live `kind = 'file'`, current, non-trashed, non-deleted by default; flags widen to thumbnails, old versions, trashed, deleted). Here it gates the counts in `query_all_tags_with_counts`; the downstack file, directory, stats, and thumbnail PRs reuse it for every listing and count.


alexfreska:
* **#772** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/772?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/772?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/sia-storage-core/src/db/operations/tags.rs
```diff
@@ -0,0 +1,1055 @@
...
+    Ok(tag)
+}
+
+/// Return the tag named `name`, creating it if absent, and bump its `usedAt` to now
+/// either way. Errors `EmptyName` on a blank name.
+pub fn get_or_create_tag(conn: &Connection, name: &str) -> Result<Tag, TagError> {
+    let trimmed = require_non_empty(name)?;
+    let now = Utc::now().timestamp_millis();
+    conn.execute(
+        "INSERT OR IGNORE INTO tags (id, name, createdAt, usedAt, system) VALUES (?1, ?2, ?3, ?3, 0)",
```
peterjan: can't we replace this with 

```
INSERT INTO tags (id, name, createdAt, usedAt, system)
VALUES (?1, ?2, ?3, ?3, 0)
ON CONFLICT(name) DO UPDATE SET usedAt = excluded.usedAt
RETURNING id, name, createdAt, usedAt, system
```

that would be cleaner overall and I think it might get rid of `TagError::GetOrCreateFailed` all together as well

crates/sia-storage-core/src/db/operations/tags.rs
```diff
@@ -0,0 +1,1055 @@
...
+}
+
+/// Tag a file by name, creating the tag if needed, and bump its `updatedAt`.
+pub fn add_tag_to_file(conn: &Connection, file_id: &str, tag_name: &str) -> Result<(), TagError> {
+    let tag = get_or_create_tag(conn, tag_name)?;
+    conn.execute(
+        "INSERT OR IGNORE INTO file_tags (fileId, tagId) VALUES (?, ?)",
+        params![file_id, tag.id],
+    )?;
+    touch_files(conn, &[file_id.to_string()])?;
```
peterjan: doesn't this mean files are touched regardless of whether the tag relation pre-existed or not? I'd check the affected row count and touch files conditional on that, e.g. `changed = conn.execute... if changed > 0 touch_files`

crates/sia-storage-core/src/db/operations/tags.rs
```diff
@@ -0,0 +1,1055 @@
...
+    // Flag carriers before deleting the links: the flag reads them through the
+    // file_tags subquery.
+    touch_files_with_tag(conn, tag_id, Utc::now().timestamp_millis())?;
+    conn.execute("DELETE FROM file_tags WHERE tagId = ?", params![tag_id])?;
+    conn.execute("DELETE FROM tags WHERE id = ?", params![tag_id])?;
+    Ok(())
+}
+
+/// Tag a file by name, creating the tag if needed, and bump its `updatedAt`.
+pub fn add_tag_to_file(conn: &Connection, file_id: &str, tag_name: &str) -> Result<(), TagError> {
```
peterjan: can these singular methods, like add tag/remove tag/ upsert object and insert log, not call the batch methods by wrapping the parameters and calling it internally, to avoid duplication the query

crates/sia-storage-core/src/db/operations/tags.rs
```diff
@@ -0,0 +1,1055 @@
...
+    Ok(())
+}
+
+/// Unlink a tag from a file and bump the file's `updatedAt`.
+pub fn remove_tag_from_file(conn: &Connection, file_id: &str, tag_id: &str) -> Result<(), DbError> {
+    conn.execute(
+        "DELETE FROM file_tags WHERE fileId = ? AND tagId = ?",
+        params![file_id, tag_id],
+    )?;
+    touch_files(conn, &[file_id.to_string()])?;
```
peterjan: same applies here, we always touch the file regardless of whether it had the tag, this one is probably less of an issue considering we probably wouldn't call the method if it didn't have the tag...

crates/sia-storage-core/src/db/operations/tags.rs
```diff
@@ -0,0 +1,1055 @@
...
+
+    let missing: Vec<String> = names
+        .iter()
+        .filter(|n| !name_to_id.contains_key(*n))
+        .cloned()
+        .collect();
+    if !missing.is_empty() {
+        let now = Utc::now().timestamp_millis();
+        let mut stmt = conn.prepare(
+            "INSERT OR IGNORE INTO tags (id, name, createdAt, usedAt, system) VALUES (?1, ?2, ?3, ?3, 0)",
```
peterjan: This is repeating the query once more, perhaps we should consider extracting the query into a const and using it in all call sites. 

```
fn ensure_tags_by_name<'a, I>(
    conn: &Connection,
    names: I,
) -> Result<HashMap<String, String>, TagError>
where
    I: IntoIterator<Item = &'a str>,
{
    let now = Utc::now().timestamp_millis();
    let mut tags = HashMap::new();
    let mut stmt = conn.prepare(UPSERT_TAG_BY_NAME_SQL)?;

    for raw_name in names {
        let name = raw_name.trim();

        if name.is_empty() || tags.contains_key(name) {
            continue;
        }

        let tag = stmt.query_row(
            params![unique_id(), name, now],
            tag_from_db_row,
        )?;

        tags.insert(tag.name, tag.id);
    }

    Ok(tags)
}
```

this way we also return a map keyed by the trimmed, canonical tag name, so the call sites should do the lookup on the trimmed tag but I think that'll end up being useful because you can get rid of the `IndexSet` seeing as `ensure_tags_by_name` now handles trimming and deduplication

peterjan (changes requested):

crates/sia-storage-core/src/db/operations/tags.rs
alexfreska: done

crates/sia-storage-core/src/db/operations/tags.rs
alexfreska: updated

crates/sia-storage-core/src/db/operations/tags.rs
alexfreska: updated

crates/sia-storage-core/src/db/operations/tags.rs
alexfreska: updated

crates/sia-storage-core/src/db/operations/tags.rs
alexfreska: Left the wrappers separate because their touch semantics differ

crates/sia-storage-core/src/db/operations/tags.rs
```diff
@@ -0,0 +1,1209 @@
...
+    pub async fn ensure_system_tags(&self) -> Result<(), DbError> {
+        self.transaction(move |c| ensure_system_tags_stmt(c)).await
+    }
+
+    /// Create a new user tag from a trimmed name. Errors `EmptyName` on a blank name and
+    /// `AlreadyExists` when a tag of that name is already present.
+    pub async fn insert_tag(&self, name: &str) -> Result<Tag, TagError> {
+        let trimmed = require_non_empty(name)?;
+        self.transaction(move |c| {
+            if read_tag_by_name(c, &trimmed)?.is_some() {
```
n8mgr: You can catch the unique constraint violation in the `INSERT` below instead of doing two queries.

crates/sia-storage-core/src/db/operations/tags.rs
```diff
@@ -0,0 +1,1209 @@
...
+            c.execute("DELETE FROM file_tags WHERE tagId = ?", params![tag_id])?;
+            c.execute("DELETE FROM tags WHERE id = ?", params![tag_id])?;
+            Ok(Ok(()))
+        })
+        .await?
+    }
+
+    /// Tag a file by name, creating the tag if needed, and bump its `updatedAt`.
+    pub async fn add_tag_to_file(&self, file_id: &str, tag_name: &str) -> Result<(), TagError> {
+        let (file_id, tag_name) = (file_id.to_string(), tag_name.to_string());
```
n8mgr: if you're going to immediately convert to an owned `String`, pass in an owned `String` instead. `to_string` on a borrow clones and that may not be necessary.

crates/sia-storage-core/src/db/operations/tags.rs
```diff
@@ -0,0 +1,1209 @@
...
+    /// Tag every file in `file_ids` by name (creating the tag once).
+    pub async fn add_tag_to_files(
+        &self,
+        file_ids: &[String],
+        tag_name: &str,
+    ) -> Result<(), TagError> {
+        if file_ids.is_empty() {
+            return Ok(());
+        }
+        let (file_ids, tag_name) = (file_ids.to_vec(), tag_name.to_string());
```
n8mgr: Same here. You immediatel convert your borrowed array to an owned `Vec`. Just take the owned `Vec`.

crates/sia-storage-core/src/db/operations/filter.rs
```diff
@@ -0,0 +1,199 @@
...
+    pub include_deleted: bool,
+}
+
+/// The active-record predicate every file, directory, and stats query funnels through. By default it
+/// scopes to the live file group: `kind = 'file' AND current = 1 AND trashedAt IS NULL AND deletedAt
+/// IS NULL`. Each `BuildRecordFilterOpts` flag widens the set (thumbnails, old versions, trashed,
+/// deleted). `table_alias` must not be `o`: the thumbnail branch's EXISTS subquery uses `o` for the
+/// original's row, and an outer `o` would self-correlate and admit no thumbnails.
+pub(crate) fn build_record_filter(table_alias: &str, opts: BuildRecordFilterOpts) -> String {
+    let a = table_alias;
```
n8mgr: Why alias this?

crates/sia-storage-core/src/db/operations/tags.rs
```diff
@@ -0,0 +1,1209 @@
...
+
+pub(crate) const SYSTEM_TAG_FAVORITES: SystemTag = SystemTag {
+    id: "sys:favorites",
+    name: "Favorites",
+};
+
+/// The built-in tags seeded into every database (currently just Favorites).
+pub(crate) const SYSTEM_TAGS: &[SystemTag] = &[SYSTEM_TAG_FAVORITES];
+
+fn require_non_empty(name: &str) -> Result<String, TagError> {
```
n8mgr: If this took an owned String, you can trim in place without the clone.

crates/sia-storage-core/src/db/operations/tags.rs
```diff
@@ -0,0 +1,1209 @@
...
+            )?;
+            Ok(Ok(tag))
+        })
+        .await?
+    }
+
+    /// Return the tag named `name`, creating it if absent, and bump its `usedAt` to now either
+    /// way. Errors `EmptyName` on a blank name.
+    pub async fn get_or_create_tag(&self, name: &str) -> Result<Tag, TagError> {
+        let name = name.to_string();
```
n8mgr: Pass in the owned string and move it instead of cloning on borrow.

crates/sia-storage-core/src/lib_utils/unique_id.rs
```diff
@@ -0,0 +1,17 @@
+/// A random unique id: a UUIDv4 in simple form (32 hex chars). Ids are opaque,
+/// uniqueness-only; nothing parses the format.
+pub fn unique_id() -> String {
+    uuid::Uuid::new_v4().simple().to_string()
```
n8mgr: Why add the UUID crate if the goal is 32 random bytes of data? Use random + hex.

crates/sia-storage-core/src/db/operations/filter.rs
alexfreska: Meh, just to be short `a` in the queries, changed to the full name.

crates/sia-storage-core/src/db/operations/tags.rs
alexfreska: updated

crates/sia-storage-core/src/db/operations/tags.rs
```diff
@@ -0,0 +1,1209 @@
...
+/// reject it later.
+pub(in crate::db) fn ensure_system_tags_stmt(conn: &Connection) -> Result<(), DbError> {
+    let now = Utc::now().timestamp_millis();
+    let mut stmt = conn.prepare(
+        "INSERT OR IGNORE INTO tags (id, name, createdAt, usedAt, system) VALUES (?1, ?2, ?3, ?3, 1)",
+    )?;
+    for tag in SYSTEM_TAGS {
+        stmt.execute(params![tag.id, tag.name, now])?;
+    }
+    Ok(())
```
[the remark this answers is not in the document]
alexfreska: system tags are seeded in the migration before any user tag exists

alexfreska:
@n8mgr updated ownership: anything that moves its arg into the transaction takes it owned now instead of borrowing and cloning, changed this for all types, string, vectors, structs. Also swapped uuid for getrandom + hex.

crates/sia-storage-core/src/lib_utils/unique_id.rs
alexfreska: swapped in getrandom + hex.

PR #773, opened 2026-07-02
feat(core): file db operations

alexfreska:
The `files` table ops: each file record's metadata plus the per-(name, directoryId) `current` pointer that ranks a file's version stack. Reads run through the shared active-set filter (current, non-trashed, non-deleted, thumbnails excluded) and a `FileQueryOpts` builder that assembles the WHERE/ORDER/LIMIT for the list and count queries. Also lands `natural_sort_key` for the `nameSortKey` column and the `FileRecord`/`FileKind`/`ThumbSize` types.

```rust
db.transaction(|tx| insert_file(tx, &row, InsertFileOptions::default())).await?;
let files = db.transaction(|tx| query_files(tx, opts)).await?;
```

What it adds, grouped by the flow that drives it:

Listing:
- `query_files`: LEFT JOIN(files, objects) list query from `FileQueryOpts`, drives the uploader's candidate scan and the library list.
- `count_files`: row count off the same builder, for the list header.
- `query_file_stats`: total bytes off the same builder, for the list header.
- `query_files_by_ids`: rows keyed by id, for batch lookups and the trash-purge batch.

Point reads:
- `read_file`: a row plus its object refs keyed by indexerURL, the main single-file read.
- `read_file_by_name`: same, keyed by name and directory.
- `read_file_by_name_in_unfiled`: same, for a name at the unfiled root.
- `read_file_by_content_hash`: same, keyed by content hash.
- `read_file_by_object_id`: same, resolving an indexer object id to its file on sync-down.
- `read_files_by_ids`: a batch of rows plus refs by id.
- `query_file_by_id`: a bare row for an id, used on the download and upload paths.

Import and edit:
- `insert_file`: create a file record.
- `create_file_with_local_object`: create a file backed by a local object.
- `insert_many_files`: bulk create.
- `update_file`: edit a file record.
- `update_file_with_local_object`: edit a file that has a local object.

Every write in this group flags the file's objects dirty for sync-up and recalcs `current`.

Sync-down apply:
- `upsert_many_files`: land file records decoded from indexer metadata during sync-down, inserting new ones and updating known ones; the update set deliberately omits `directoryId`/`localId`/`addedAt`/`deletedAt`/`lostReason` so a re-sync can't clobber local-only state.
- `tombstone_files`: apply a delete, `AlreadyRemote` on sync-down so it won't re-flag what the indexer already removed, `Flag` for a local delete.
- `query_files_by_object_ids`: map indexer object ids back to file rows.

Move and rename:
- `move_all_file_versions`: move a file's whole version stack under staggered decreasing timestamps, so version order holds while every row bumps to sync up.
- `rename_all_file_versions`: same for a rename.
- `move_files_all_versions`: bulk-move many stacks under one global newest-first stamp, so a cross-stack same-name merge picks the globally newest version as `current`.

Delete:
- `delete_file_by_id`: hard-delete one file.
- `delete_many_files_by_ids`: hard-delete a batch.
- `delete_all_files`: hard-delete every file.
- `delete_file_and_thumbnails`: hard-delete a file plus its thumbnail rows and recalc `current`, for the import screen's remove-unavailable action.
- `delete_files_and_thumbnails`: same for a batch.

Current pointer:
- `recalculate_current_for_group`: re-rank one (name, directoryId) group, the primitive every write and the move/rename ops call.
- `recalculate_current_for_groups`: the same across many groups.
- `recalculate_current_for_file_ids`: set-based bulk recalc on the sync-down hot path.

Versions and status:
- `query_file_versions`: a file's version history, for the version-history sheet and the rename/move ops.
- `query_uploaded_file_ids`: ids of fully-uploaded files, for the `ls` uploaded badge.
- `count_unuploaded_files` / `query_unuploaded_files`: count and rows for the not-uploaded screen.
- `count_lost_files` / `query_lost_file_stats` / `query_lost_files`: count, bytes, and rows for the lost-files screen.
- `delete_lost_files_and_thumbnails`: remove lost files plus their thumbnails.
- `query_files_by_local_ids`: rows by local id.
- `query_current_files_by_names_in_directory`: current rows for a set of names in a directory.
- `query_files_by_content_hashes`: rows by content hash.
- `query_directory_ids_for_files`: directory ids for a set of files.
- `query_active_file_summaries`: active-set summaries.

The last five serve import dedupe and directory lookups.

- `read_file_with_objects`: a file row plus its full objects with slabs, for the download/upload path.
- `query_local_only_files` / `count_local_files` / `query_local_file_stats`: rows, count, and bytes for a local-only file listing.
- `read_files_by_local_ids` / `read_current_files_by_names_in_directory` / `read_files_by_content_hashes`: the row-plus-refs variants of the by-local-id, by-name, and by-content-hash lookups.
- `delete_thumbnails_by_file_id`: drop a file's thumbnail rows directly.

These round out the op surface so later PRs add callers without touching this file.


alexfreska:
* **#773** <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/773?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/SiaFoundation/sia-storage-app/773?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/sia-storage-core/src/types/files.rs
```diff
@@ -0,0 +1,169 @@
...
+    pub added_at: i64,
+    pub deleted_at: Option<i64>,
+    pub lost_reason: Option<String>,
+}
+
+#[cfg(test)]
+impl FileRecordRow {
+    /// Test fixture: constructs a minimal valid `file`-kind row with
+    /// deterministic defaults; chain the setters below to vary fields.
+    pub fn test(id: impl Into<String>) -> Self {
```
n8mgr: pub?

crates/sia-storage-core/src/types/files.rs
```diff
@@ -0,0 +1,169 @@
...
+            updated_at: 1000,
+            local_id: None,
+            added_at: 1000,
+            deleted_at: None,
+            lost_reason: None,
+            id,
+        }
+    }
+
+    pub fn name(mut self, v: impl Into<String>) -> Self {
```
n8mgr: Why the setter? The field is public.

crates/sia-storage-core/src/types/files.rs
```diff
@@ -0,0 +1,169 @@
...
+        self.hash = v.into();
+        self
+    }
+    /// The import-placeholder case: a row whose content hash is not yet known.
+    pub fn empty_hash(mut self) -> Self {
+        self.hash = String::new();
+        self
+    }
+    /// Mark this row as a thumbnail OF `file_id` at `size` (sets kind = Thumb).
+    pub fn thumb_for(mut self, file_id: impl Into<String>, size: ThumbSize) -> Self {
```
n8mgr: The setter for this makes more sense, but you can also create a "bad" record by updating the fields outside of it. The publicly reachable API should try to protect against unsafe variants as much as possible.

crates/sia-storage-core/src/db/operations/files.rs
```diff
@@ -0,0 +1,2534 @@
...
+            let objs = objects_by_id.remove(&id);
+            transform_row(row, objs)
+        })
+        .collect()
+}
+
+/// Builds the domain `FileRecord` from a row + its object refs, keying objects by indexerURL.
+/// The objects are slim ref stubs, NO slabs; use `read_file_with_objects` when the full objects
+/// are needed.
+pub fn transform_row(row: FileRecordRow, objects: Option<Vec<LocalObjectRef>>) -> FileRecord {
```
n8mgr: pub?

crates/sia-storage-core/src/db/operations/files.rs
```diff
@@ -0,0 +1,2534 @@
...
+        size: r.get("size")?,
+    })
+}
+
+/// Predicate behind all three "lost file" queries: lostReason set OR (non-empty hash + no
+/// objects on this indexer + no local fs entry). An empty-hash row is never lost via the second
+/// rule.
+const LOST_PREDICATE_SQL: &str = r"(
+       f.lostReason IS NOT NULL
+       OR (NOT EXISTS (SELECT 1 FROM objects s WHERE s.fileId = f.id AND s.indexerURL = ?)
```
n8mgr: A random SQL param in here seems dangerous. Easy way to introduce a bug if another parameter ever happens to make it before this predicate.

crates/sia-storage-core/src/db/operations/files.rs
```diff
@@ -0,0 +1,2534 @@
...
+pub struct FileUpdate {
+    pub id: String,
+    pub name: Option<String>,
+    pub type_: Option<String>,
+    pub kind: Option<FileKind>,
+    pub size: Option<u64>,
+    pub hash: Option<String>,
+    pub created_at: Option<i64>,
+    pub updated_at: Option<i64>,
+    pub thumb_for_id: Option<Option<String>>,
```
n8mgr: The nested `Option`'s are a bit gross and unclear on whats happening. Could be replaced by an enum? It is not really different, but it is more clear. It took me a minute to understand what you were doing with the options.

```rs
#[derive(Debug, Clone, Default, PartialEq, Eq)]
pub enum FieldUpdate<T> {
    #[default]
    Unchanged,
    Changed(T)
}
```

crates/sia-storage-core/src/db/operations/files.rs
```diff
@@ -0,0 +1,2534 @@
...
+    let mut stmt = conn.prepare(
+        "SELECT DISTINCT directoryId FROM files WHERE id IN rarray(?) AND directoryId IS NOT NULL",
+    )?;
+    let out = stmt
+        .query_map([sql::id_array(file_ids)], |r| r.get(0))?
+        .collect::<rusqlite::Result<Vec<String>>>()?;
+    Ok(out)
+}
+
+pub(in crate::db) fn upsert_many_files_stmt(
```
peterjan, line 937: Why doesn't `upsert_many_files_stmt` have as `update_file_stmt` to fetch the old group? `recalc_if_needed` will recalculate it from post update rows right? This is a little hard to follow for me of course but I think it can leave a file stranded with no current=1 on a sync down rename

crates/sia-storage-core/src/db/operations/files.rs
```diff
@@ -0,0 +1,2534 @@
...
+                  AND {}",
+                LOST_PREDICATE_SQL
+            );
+            let ids: Vec<String> = {
+                let mut stmt = c.prepare(&sql)?;
+                stmt.query_map(params![indexer_url], |r| r.get(0))?
+                    .collect::<rusqlite::Result<Vec<String>>>()?
+            };
+            let total = ids.len() as u64;
+            delete_files_and_thumbnails_stmt(c, &ids)?;
```
peterjan, line 1545: What happens if I upload a file on indexer A, then switch to indexer B, then we run this with the new indexer URL, wouldn't the file be counted as lost, get deleted and it would delete the object as well, even though it still exists on A?

crates/sia-storage-core/src/db/operations/files.rs
```diff
@@ -0,0 +1,2534 @@
...
+            transform_row(row, objs)
+        })
+        .collect()
+}
+
+/// Builds the domain `FileRecord` from a row + its object refs, keying objects by indexerURL.
+/// The objects are slim ref stubs, NO slabs; use `read_file_with_objects` when the full objects
+/// are needed.
+pub fn transform_row(row: FileRecordRow, objects: Option<Vec<LocalObjectRef>>) -> FileRecord {
+    let mut objects_map: HashMap<String, crate::types::files::LocalObjectRefDto> = HashMap::new();
```
peterjan, line 209: Is it legal to key by indexer URL? I'm not too familiar with the design but this feels like it could squash rows unexpectedly? I think there also might be some multi-indexer scenarios missing in testing, feels similar to my question on `delete_lost_files_and_thumbnails`

crates/sia-storage-core/src/db/operations/files.rs
```diff
@@ -0,0 +1,2534 @@
...
+            params.push(Value::Integer(i64::from(n)));
+            " LIMIT ?".to_string()
+        }
+        None => String::new(),
+    };
+    (where_sql, params, order_expr, limit_expr)
+}
+
+/// Run a single-group aggregate over `files` with the given SELECT `projection`.
+/// The `ORDER BY {order_expr}` is inert on a single-group aggregate (one row, no GROUP BY), but
```
peterjan: might leave a note on how limit is handled with aggregate queries too

crates/sia-storage-core/src/db/operations/files.rs
```diff
@@ -0,0 +1,2534 @@
...
+        assignments.push(("updatedAt", Value::Integer(u)));
+    }
+    if let Some(name) = &update.name {
+        assignments.push(("nameSortKey", natural_sort_key(Some(name)).into()));
+    }
+    if !options.include_updated_at {
+        assignments.push(("updatedAt", Value::Integer(Utc::now().timestamp_millis())));
+    }
+
+    let needs_recalc = options.current_recalc == CurrentRecalc::Run
```
peterjan, line 829: wait don't we bump updated_at on every options update (unless `include_updated_at` is set and we use that value)? and then we don't recalculate, isn't it possible for a non-current row to have the highest updatedAt so an older version might rank higher/first?

crates/sia-storage-core/src/db/operations/files.rs
alexfreska: switched it to a named :indexer_url bind

crates/sia-storage-core/src/db/operations/files.rs
alexfreska: agreed, updated to this pattern.

crates/sia-storage-core/src/db/operations/files.rs
alexfreska: cleaned this up by splitting the query helpers

crates/sia-storage-core/src/db/operations/files.rs
alexfreska: Yeah, one object per indexer is intended here, if a file ever had two on the same indexer (rare edge case) it doesn't matter which one we use here. Object deletes go by fileId anyway (plus the FK cascade), so both rows get cleaned up regardless.

crates/sia-storage-core/src/db/operations/files.rs
```diff
@@ -0,0 +1,2641 @@
...
+                    updated_at: o.updated_at.timestamp_millis(),
+                },
+            );
+        }
+    }
+    FileRecord {
+        row,
+        objects: objects_map,
+    }
+}
```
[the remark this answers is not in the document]
alexfreska: Yeah, one object per indexer is intended here, if a file ever had two on the same indexer (rare edge case) it doesn't matter which one we use here. Object deletes go by fileId anyway (plus the FK cascade), so both rows get cleaned up regardless.

crates/sia-storage-core/src/db/operations/files.rs
alexfreska: Updated the comments to explain this.

crates/sia-storage-core/src/db/operations/files.rs
alexfreska: yeah, good catch, fixed. current is the newest updatedAt in the group, but the recalc only fired when the caller passed updatedAt, not on the auto-bump, so it could lag. keys off the actual updatedAt write now.

crates/sia-storage-core/src/db/operations/files.rs
alexfreska: Yeah, good catch, added a guard and test to ensure we only delete files with no object on any indexer.

crates/sia-storage-core/src/db/operations/files.rs
alexfreska: Good catch. fixed in both modes: under Run it now grabs the rows' (name, directoryId) groups before the upsert and recalcs them after, same as update_file_stmt, and the sync-down batch (which passes Skip and owns its consolidated recalc) captures them there. Tests cover both.

crates/sia-storage-core/src/types/files.rs
alexfreska: These fixtures get used by tests all over the crate, and eventually by the integration tests crate too, so I changed the gate to `cfg(any(test, feature = "test-fixtures"))`. That's also why they're pub.

crates/sia-storage-core/src/types/files.rs
alexfreska: It just makes tests (used in many outside this module) way easier to write. eg:

`let thumb = FileRecordRow::test("t1").thumb_for("photo", ThumbSize::S64).size(10);`

defaults for everything, chain setters.

n8mgr: Use Default or init it normally with every field. This is working around a Rust protection that's honestly useful.

crates/sia-storage-core/src/types/files.rs
alexfreska: These factories are intended only for test convenience. The type itself is intended to be used publicly and the db layer enforces the invariants. Let me know if this is the right approach for that.

n8mgr: Yeah, better to not add public api surface for minor test convenience

alexfreska: Ok I'll remove.

crates/sia-storage-core/src/types/files.rs
alexfreska: Ok how about just have `test(id)` and `test_thumb(id, parent, size)` for valid default rows, and drop all the field setters. Overrides become normal struct update syntax off the base.

before:

```rust
￼let v1 = FileRecordRow::test("v1").name("a.txt").updated_at(100);
 let thumb = FileRecordRow::test("t1").thumb_for("photo", ThumbSize::S64).size(10);￼
```

after:

```rust
let v1 = FileRecordRow {  
  name: "a.txt".into(),  
  updated_at: 100,  
  ..FileRecordRow::test("v1")  
};  
let thumb = FileRecordRow {  
  size: 10,  
  ..FileRecordRow::test_thumb("t1", "photo", ThumbSize::S64)  
};  
```

same defaults in one place, but the overrides are plain field writes instead of a setter API.

alexfreska: Never mind, totally removed in favor of factory per test file, tailored to what that file seeds.

alexfreska:
@n8mgr totally removed the factory helpers, @peterjan fixed the issues you found and added tests.

crates/sia-storage-core/src/db/operations/files.rs
```diff
@@ -0,0 +1,2693 @@
...
+    pub order: Option<FileQueryOrder>,
+    pub pinned: Option<FileQueryPinned>,
+    pub file_exists_locally: Option<bool>,
+    pub exclude_ids: Option<Vec<String>>,
+    pub include_thumbnails: bool,
+    pub include_old_versions: bool,
+    pub include_trashed: bool,
+    pub include_deleted: bool,
+    pub hash_empty: bool,
+    pub hash_not_empty: bool,
```
n8mgr: What happens if `hash_empty = true` and `hash_not_empty = true`?

alexfreska: matches nothing, fixed it by collapsing into a single hash: Option\<HashFilter>

